PR #17, opened 2021-08-09
Hocon impl

Nightcro:

## Other Changes

Added test config for integration tests.
Added hocon configuration alternative for paxos and ble components.


Nightcro:
I did not create a configuration file for ble and paxos in the tests as I was not sure if I should create different hocon configurations for each component as we provide the pid through the hocon configuration.
There are also some parameters for ble and paxos that can not be passed in a hocon configuration.

src/key_values.rs
```diff
@@ -0,0 +1,10 @@
+/// The identifier for the configuration that this Omni-Paxos replica is part of.
```
haraldng, line 1: Let's not have this file under `/src`. Can you create a new module `utils` and put it there? Also the filename can be more clear.

I think something like:  `omnipaxos/utils/hocon_kv.rs` would be good.

src/leader_election.rs
```diff
@@ -139,6 +141,27 @@ pub mod ballot_leader_election {
...
+            &self,
+            cfg: &Hocon,
+            peers: Vec<u64>,
+            initial_leader: Option<Leader<Ballot>>,
+        ) -> BallotLeaderElection {
+            BallotLeaderElection::with(
+                peers,
+                cfg[PID].as_i64().unwrap_or_default() as u64,
+                cfg[HB_DELAY].as_i64().unwrap_or_default() as u64,
+                cfg[INCREMENT_DELAY].as_i64().unwrap_or_default() as u64,
```
haraldng, line 159: Can you put these into a private function? And instead of `unwrap_or_default()` use `expect()` with a message that lets the user know what value failed to be parsed. For example: "Failed to load PID"

src/paxos.rs
```diff
@@ -140,6 +143,28 @@ where
...
+    pub fn with_hocon(
+        &self,
+        cfg: &Hocon,
+        peers: Vec<u64>,
+        storage: Storage<R, S, P>,
+        skip_prepare_use_leader: Option<Leader<R>>,
+    ) -> Paxos<R, S, P> {
+        Paxos::<R, S, P>::with(
+            cfg[CONFIG_ID].as_i64().unwrap_or_default() as u32,
+            cfg[PID].as_i64().unwrap_or_default() as u64,
```
haraldng, line 161: Same here: put into private function and use expect().

tests/data/basic.conf
```diff
@@ -0,0 +1,25 @@
+ble_test {
```
haraldng, line 1: instead of `/data` let's use `/config`

haraldng (changes requested):

haraldng (approved):

PR #20, opened 2021-08-18
Add garbage collection

Nightcro:

## Issues

Added garbage collection.
Added 1 test for checking.

At the moment the implementation is barebone. The user is not aware if the gc succeeded or not. The replicas do not answer back to the leader.

### DIFF
--- a/src/paxos.rs
+++ b/src/paxos.rs
@@ -7,7 +7,7 @@ use crate::{
     util::PromiseMetaData,
 };
 use hocon::Hocon;
-use slog::{crit, debug, info, trace, Logger};
+use slog::{crit, debug, info, trace, warn, Logger};
 use std::{fmt::Debug, sync::Arc};
 
 const BUFFER_SIZE: usize = 100000;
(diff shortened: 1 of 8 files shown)


src/storage.rs
```diff
@@ -96,6 +96,9 @@ where
     /// Returns true if the log contains a StopSign or a StopSign already has been decided.
     /// Note that the log could have a StopSign that later gets overwritten, and thus this function might first return true and later false.
     fn stopped(&self) -> bool;
+
+    /// Starts the garbage collection process to free storage.
```
haraldng, line 100: This function does not start the garbage collection process (which is done in the `OmniPaxos` struct), it is rather just removing the first `idx` elements of the log.

```suggestion
    /// Removes elements up to the given [`idx`] from storage.
```

src/storage.rs
```diff
@@ -123,6 +126,12 @@ where
 
     /// Returns the round that has been promised.
     fn get_promise(&self) -> R;
+
+    /// Sets the garbage collector index.
```
haraldng, line 130:
```suggestion
    /// Sets the garbage collected index.
```

src/storage.rs
```diff
@@ -123,6 +126,12 @@ where
...
+    fn set_gc_idx(&mut self, index: u64);
+
+    /// Returns the garbage collector index.
```
haraldng, line 133:
```suggestion
    /// Returns the garbage collected index.
```

tests/config/test.conf
```diff
@@ -22,4 +22,14 @@ proposal_test {
     num_nodes: 6
```
haraldng, line 22: Why is this 6? Use uneven numbers

tests/config/test.conf
```diff
@@ -22,4 +22,14 @@ proposal_test {
...
     ble_hb_delay: 5
     increment_delay: 2
-}
\ No newline at end of file
+}
+
+gc_test {
+    wait_timeout: 2 s
+    num_threads: 8
+    num_nodes: 11
```
haraldng, line 30: I don't think 11 nodes are needed, consensus clusters are usually at most 7 nodes.

tests/util.rs
```diff
@@ -393,6 +393,18 @@ pub mod omnireplica {
...
+        }
+
+        pub fn garbage_collect_by_index(&mut self, index: u64) {
+            self.paxos.garbage_collect_by_index(index)
+        }
+
+        pub fn get_garbage_collector_index(&self) -> u64 {
+            self.paxos.get_gc_from_storage()
+        }
+
```
haraldng, line 407: Do we really need these functions if their body is one line anyway? :D

src/paxos.rs
```diff
@@ -189,6 +191,72 @@ where
...
     }
 
+    /// Initiates the garbage collection process using the provided index.
+    /// # Arguments
+    /// * `index` - Deletes all entries up to [`index`]
+    pub fn garbage_collect_by_index(&mut self, index: u64) {
+        self.gc_propose(Some(index));
+    }
+
+    /// Initiates the garbage collection process using the minimum last accepted sequence as the [`index`].
```
haraldng, line 201:
```suggestion
    /// Initiates the garbage collection process using the minimum index accepted by **ALL** servers as the [`index`].
```

src/paxos.rs
```diff
@@ -189,6 +191,72 @@ where
...
+    pub fn garbage_collect(&mut self) {
+        self.gc_propose(None);
+    }
+
+    /// Return garbage collection index from storage.
+    pub fn get_gc_from_storage(&self) -> u64 {
+        self.storage.get_gc_idx()
+    }
+
+    fn gc_propose(&mut self, index: Option<u64>) {
```
haraldng, line 211: This function appear unnecessary... just put its body into `garbage_collect`

src/paxos.rs
```diff
@@ -443,6 +522,19 @@ where
         }
     }
 
+    fn forward_gc_proposal(&mut self, index: Option<u64>) {
+        if self.leader > 0 && self.leader != self.pid {
+            trace!(
+                self.logger,
+                "Forwarding gc proposal to Leader {}",
```
haraldng, line 529: "gc request" and the index of it

src/paxos.rs
```diff
@@ -189,6 +191,72 @@ where
...
+            );
+            return;
+        }
+
+        trace!(self.logger, "Garbage Collection index: {:?}", index);
+
+        self.cached_gc_index = index.unwrap();
+
+        self.storage.set_gc_idx(index.unwrap());
+        self.storage.garbage_collect(index.unwrap());
```
haraldng, line 257: we **must** call `self.storage.garbage_collect()` first. We don't want to first set the index and then crash and when we recover we will have the updated gc_index but the entries are still in the log.

src/storage.rs
```diff
@@ -308,6 +317,28 @@ where
...
+                s.garbage_collect(idx);
+                self.paxos_state.set_gc_idx(idx);
+            }
+            PaxosSequence::Stopped(_) => {} // todo what to do when paxos is stopped?
+            _ => panic!("Got unexpected intermediate PaxosSequence::None in stopped()"),
+        }
+    }
+
+    /// Sets the garbage collector index in storage.
+    pub fn set_gc_idx(&mut self, index: u64) {
```
haraldng, line 334: Remove this function. The storage should always perform the gc and then set the gc_idx internally. A user should not be able to only set the gc_idx without actually garbage collecting.

haraldng (changes requested):

tests/util.rs
Nightcro: I can delete the first one, but the other two I need for the tests

src/paxos.rs
```diff
@@ -189,6 +191,83 @@ where
...
+                decided_len
+            );
+            return;
+        }
+
+        trace!(self.logger, "Garbage Collection index: {:?}", index);
+
+        self.cached_gc_index = index;
+
+        self.storage.garbage_collect(index);
```
haraldng, line 268: Please put `self.storage.garbage_collect(index)` before `self.cached_gc_index = index`

src/storage.rs
```diff
@@ -308,6 +317,23 @@ where
...
     }
+
+    /// Removes elements up to the given [`idx`] from storage.
+    pub fn garbage_collect(&mut self, idx: u64) {
+        match self.sequence {
+            PaxosSequence::Active(ref mut s) => {
+                s.garbage_collect(idx - self.paxos_state.get_gc_idx());
+                self.paxos_state.set_gc_idx(idx);
+            }
+            PaxosSequence::Stopped(_) => {} // todo what to do when paxos is stopped?
```
haraldng, line 328: If we are stopped, we can carbage collect too. Here, I think we can actually use the OR `|` in pattern matching :D 

```suggestion
            PaxosSequence::Active(ref mut s) | PaxosSequence::Stopped(_) => {
                s.garbage_collect(idx - self.paxos_state.get_gc_idx());
                self.paxos_state.set_gc_idx(idx);
            }
```

src/storage.rs
```diff
@@ -393,6 +423,8 @@ pub mod memory_storage {
         acc_round: R,
         /// Length of the decided sequence.
         ld: u64,
+        /// Garbage collector index.
```
haraldng, line 426:
```suggestion
        /// Garbage collected index.
```

haraldng (changes requested):
Nice! Just one more thing apart from the comments: Can we extend the gc_test so we perform multiple garbage collections, one after the other.

haraldng (approved):

PR #45, opened 2022-03-26
fix: send stopsign at end of log

Baawa:

- [] You ran `rustfmt` on the code base before submitting (on latest nightly with rustfmt support)

## Issues

The read function would not return the StopSign at the end of the log, I think this is because of a faulty comparison between the suffix_idx and log_len.

I believe the suffix_idx should always be larger than the log_len when we have a StopSign, and by forcing the suffix_idx to be equal to the log_len, we never get the StopSign.

### DIFF
--- a/omnipaxos_core/src/sequence_paxos.rs
+++ b/omnipaxos_core/src/sequence_paxos.rs
@@ -263,9 +263,9 @@ where
         } else {
             let suffix_idx = idx - compacted_idx;
             let log_len = self.storage.get_log_len();
-            if suffix_idx >= log_len {
+            if suffix_idx > log_len {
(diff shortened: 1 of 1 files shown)


haraldng:
I think the `read() `is correct, but it is instead the `sequence_paxos::stopped()` that might be confusing. It returns the accepted **but not necessarily decided** StopSign. 

Probably in your case, `stopped()` returned only the accepted StopSign but it was not decided yet. So when [you tried to call read()](https://github.com/Baawa/omnichiselstore/blob/2cdc710038dce95fc0ace028a44590222669c6d9/src/server.rs#L381) with the decided index it did not return the StopSign.

I'll close this PR and instead fix the `stopped()` function.

PR #48, opened 2022-07-25
Persistent storage for log entries and state

### LINKED ISSUE #8: Create a type alias for node id and configuration id
Instead of u64 and u32 defined here: https://github.com/haraldng/omnipaxos/blob/master/omnipaxos_core/src/sequence_paxos.rs#L30-#L32.

### LINKED ISSUE #50: Create Builder for SequencePaxos and BLE from configs
Have a `build()` function in `SequencePaxosConfig` that does some basic asserts e.g. pid != 0 and returns `SequencePaxos`. Same for BLE.

AbyelT:

- [X] You ran `rustfmt` on the code base before submitting (on latest nightly with rustfmt support)
- [X] You reference which issue is being closed in the PR text (if applicable)
- [X] You updated the OmniPaxos book (if applicable)

## Issues

Fix #50, Fix #8 

## Breaking Changes

* All storage related items and types have been moved to the _omnipaxos_storage_ package.
* The return type of logged entries in Storage trait has changed from Array slice to Vector.  SequencePaxos.rs, Storage.rs and util.rs are modified for this return type.

## Other Changes

* Created package `omnipaxos_storage` for handling in-memory and on-disk storage.
* Added dependencies for persistent storage and data serialization.
* Implemented `PersistentStorage` for persistent storage of logged entries and replica state.
* Serialization/de-serialization of log entries, and replica state.
* Implemented `PersistentStorageConfig` for configuring path and storage options.
* Added `StorageType` Enum that lets users choose between `PersistentStorage` or `MemoryStorage` during tests. This can be set at config/test.conf in the 'storage_type' field
* Added functionality to clean persistent storage after each test
* Added `Recovery_test` for testing persistent storage and fail recovery


omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+                .read(from, ReadLimit::default())
+                .unwrap_or(MessageBuf::default());
+            let mut entries = vec![];
+            let zero_idx_to = to - from;
+            for (idx, msg) in buffer.iter().enumerate() {
+                if (idx as u64) >= zero_idx_to {
+                    break;
+                }
+                entries.push(FromBytes::read_from(msg.payload()).unwrap());
+            }
```
haraldng, line 135: This loop looks a bit confusing/over-complicated. You can probably do something like this:
```
let iter = buffer.iter();
for _ in from..to {
    let msg = iter.next();
    entries.push(FromBytes::read_from(msg.payload()).unwrap());
} 
```

omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        fn get_suffix(&self, from: u64) -> Vec<T> {
+            //println!("get suffix!");
+            self.get_entries(from, self.get_log_len())
+        }
+
+        fn get_promise(&self) -> Ballot {
+            match self.db.get(b"n_prom") {
+                Ok(Some(mut value)) => {
+                    let prom_bytes: &mut [u8] = &mut value;
```
haraldng, line 154: Do you really need `mut` here? I think it should be possible to deserialize from immutable data.

omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+                    FromBytes::read_from(prom_bytes).unwrap()
+                }
+                Ok(None) => Ballot::default(),
+                Err(_e) => Ballot::default(),
+            }
+        }
+
+        fn set_promise(&mut self, n_prom: Ballot) {
+            let prom_bytes = AsBytes::as_bytes(&n_prom);
+            self.db.put(b"n_prom", prom_bytes).unwrap()
```
haraldng, line 164: Don't use string as the key in rocksdb. Use `u8` and then define and use constants instead e.g. `const NPROM: u8 = 1` etc for all keys.

omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        fn get_decided_idx(&self) -> u64 {
+            match self.db.get(b"ld") {
+                Ok(Some(value)) => {
+                    let ld_bytes: &[u8] = &value;
+                    FromBytes::read_from(ld_bytes).unwrap()
+                }
+                Ok(None) => 0,
+                Err(_e) => 0,
```
haraldng, line 174: Panic on all errors

omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+        /// path to commitlog, remove when commitlog is no longer deleted here
+        c_log_path: String,
+        /// a struct for accessing local RocksDB database
+        db: DB,
+        /// Garbage collected index.
+        trimmed_idx: u64,
+        /// Stored snapshot
+        snapshot: Option<S>,
+        /// Stored StopSign
+        stopsign: Option<StopSignEntry>,
```
haraldng, line 37: These are still in memory... not sure if you can store them in rocks but they need to be persisted too.

omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        fn get_stopsign(&self) -> Option<StopSignEntry> {
+            self.stopsign.clone()
+        }
+
+        fn trim(&mut self, trimmed_idx: u64) {
+            //println!("TRIM!");
+            let mut trimmed_log: Vec<T> = self.get_entries(0, self.c_log.next_offset()); // get the entire log, drain it until trimmed_idx
+                                                                                         // println!("the trimmed log before drain {:?}", trimmed_log);
+            trimmed_log.drain(0..trimmed_idx as usize);
```
haraldng, line 211: If the log is removed, do we still need to drain?

omnipaxos_storage/src/memory.rs
```diff
@@ -0,0 +1,374 @@
...
+            // println!("the trimmed log after drain {:?}", trimmed_log);
+            let _ = std::fs::remove_dir_all(&self.c_log_path); // remove old log
```
haraldng, line 213: Since this is a work around due to some strange stuff in commitlog, can you please add a TODO comment and create an issue later (when this PR is merged) so we don't lose track of this?

.gitignore
```diff
@@ -1,6 +1,10 @@
 /docs/book/*
 /target
+/log
 Cargo.lock
+omnipaxos_core/rocksDB/
+omnipaxos_core/commitlog/
+clean_test.sh
```
haraldng, line 7: Would it be possible to add the functionalities of this script into the tests itself? Doesn't make sense to have a bunch of garbage after each test.

examples/tutorial/core.rs
```diff
@@ -53,11 +57,13 @@ fn main() {
     sp_config.set_pid(my_pid);
     sp_config.set_peers(my_peers.clone());
 
+    //todo: hardcoded test for memory and persistent storage, remove later
```
haraldng, line 60:
```suggestion
```

haraldng (changes requested):
Good job! Some changes must still be made though: some variables must still be persisted, and if we fail to read/write the state or the log, it should panic and not just fail silently.

omnipaxos_core/tests/consensus_test.rs
```diff
@@ -73,7 +81,7 @@ fn read_test() {
 
     let exp_snapshot = LatestValue::create(snapshotted);
 
-    let mut mem_storage = MemoryStorage::<Value, LatestValue>::default();
+    let mut mem_storage = PersistentState::<Value, LatestValue>::with("read_test");
```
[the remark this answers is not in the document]
AbyelT: Ok, i'll do this with an TextConfig item to read from the configuration file. The StorageType enum can then use the config to generate a desired storage.

omnipaxos_core/Cargo.toml
```diff
@@ -3,19 +3,20 @@ name = "omnipaxos_core"
...
 edition = "2018"
-
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 slog = { version = "2.7.0", optional = true }
 slog-term = { version = "2.9.0", optional = true }
 slog-async = { version = "2.7.0", optional = true }
 hocon = { version = "0.9.0", optional = true }
+zerocopy = { version = "0.6.1", optional = false } # todo: zerocopy needed for KeyValue struct
```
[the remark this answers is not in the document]
AbyelT: Seems it was also needed for Ballot in ballot_leader_election

``#[derive(Clone, Copy, Eq, Debug, Default, Ord, PartialOrd, PartialEq, **FromBytes**, **AsBytes**)]
pub struct Ballot {
..
``

examples/tutorial/core.rs
```diff
@@ -1,28 +1,32 @@
...
+    memory_storage::MemoryStorage, persistent_storage::PersistentState,
+};
 use std::collections::HashMap;
+use zerocopy::{AsBytes, FromBytes};
 
-#[derive(Clone, Debug)]
+#[repr(packed)]
+#[derive(Copy, Clone, Debug, FromBytes, AsBytes)]
 pub struct KeyValue {
-    pub key: String,
```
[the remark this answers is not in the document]
AbyelT: This was a nested problem that came from deriving `FromBytes` and `AsBytes` structs, due to PersistentState requiring it on the KeyValue struct.

I see now that MemoryStorage is used in the core example so i will reverse this, this might be new issue for PersistentStorage i'll have to fix later.

omnipaxos_core/src/ballot_leader_election.rs
```diff
@@ -11,9 +11,11 @@ use hocon::Hocon;
 use messages::{BLEMessage, HeartbeatMsg, HeartbeatReply, HeartbeatRequest};
 #[cfg(feature = "logging")]
 use slog::{debug, info, trace, warn, Logger};
+use zerocopy::{AsBytes, FromBytes};
 
 /// Used to define an epoch
-#[derive(Clone, Copy, Eq, Debug, Default, Ord, PartialOrd, PartialEq)]
+#[repr(packed)]
```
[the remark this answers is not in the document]
AbyelT: The `AsBytes` trait required the struct to be packed.

omnipaxos_core/tests/test_config.rs
```diff
@@ -32,6 +33,9 @@ impl TestConfig {
             num_proposals: cfg["num_proposals"].as_i64().unwrap_or_default() as u64,
             num_elections: cfg["num_elections"].as_i64().unwrap_or_default() as u64,
             gc_idx: cfg["gc_idx"].as_i64().unwrap_or_default() as u64,
+            storage_type: cfg["storage_type"]
+                .as_string()
+                .unwrap_or("Memory".to_string()),
```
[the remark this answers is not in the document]
AbyelT: If its acceptable i moved the pattern matching to the StorageType implementation:

```
impl<T, S> StorageType<T, S>
    where
        T: Entry,
        S: Snapshot<T>,
    {
        pub fn with(storage_type: &str, pid: u64) -> Self {
            match storage_type {
                "Persistent" => StorageType::Persistent(PersistentStorage::with(&pid.to_string())), // Persistent storage
                "Memory" => StorageType::Memory(MemoryStorage::default()), // Memory storage (default)
                _ => panic!()
            }
        }
    }
```

.gitignore
AbyelT: Should be possible, the script only removes the `commitlog` and `rocksDB` directories before running `cargo test.` This can be done at the last test

omnipaxos_storage/src/memory.rs
AbyelT: A small miss, this line removes the need for drain :slightly_smiling_face: 

`let trimmed_log: Vec<T> = self.get_entries(trimmed_idx, self.c_log.next_offset());`

.gitignore
haraldng: No, it should be done at every test that uses PersistentStorage because we might not always run all tests.

omnipaxos_core/tests/test_config.rs
```diff
@@ -13,6 +13,7 @@ pub struct TestConfig {
     pub num_proposals: u64,
     pub num_elections: u64,
     pub gc_idx: u64,
+    pub storage_type: String,
```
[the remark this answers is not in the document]
AbyelT: Created the StorageTypeSelector enum for deciding storage type based on the hocon file. It will pattern match for either "Memory", "Storage" or panic depending on the input.

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
     where
-        T: Entry + AsBytes + FromBytes,
-        S: Snapshot<T>,
+        T: Entry + Serialize + for<'a> Deserialize<'a>,
+        S: Snapshot<T> + Serialize + for<'a> Deserialize<'a>,
     {
         fn append_entry(&mut self, entry: T) -> u64 {
-            //println!("append entry {:?}", entry);
-            let entry_bytes = AsBytes::as_bytes(&entry);
+            let entry_bytes = bincode::serialize(&entry).unwrap();
```
haraldng, line 209: (almost) Always use `expect()` with some message instead of `unwrap()`.

```suggestion
            let entry_bytes = bincode::serialize(&entry).expect("Failed to serialize");
```

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
             match self.c_log.append_msg(entry_bytes) {
                 Ok(x) => x,
-                Err(_e) => 0,
+                Err(_e) => panic!(),
             }
```
haraldng, line 213: No point having a panic without a message. Use `expect` instead

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
-        S: Snapshot<T>,
+        T: Entry + Serialize + for<'a> Deserialize<'a>,
+        S: Snapshot<T> + Serialize + for<'a> Deserialize<'a>,
     {
         fn append_entry(&mut self, entry: T) -> u64 {
-            //println!("append entry {:?}", entry);
-            let entry_bytes = AsBytes::as_bytes(&entry);
+            let entry_bytes = bincode::serialize(&entry).unwrap();
             match self.c_log.append_msg(entry_bytes) {
                 Ok(x) => x,
```
haraldng, line 211: I think this is incorrect. The Storage trait in omnipaxos is supposed to return the log length, while [commitlog](https://docs.rs/commitlog/latest/commitlog/index.html#example) returns the offset it just appended. Not sure if it might be related to the truncate problem too?

```suggestion
                Ok(x) => x + 1,
```

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
 
         fn append_entries(&mut self, entries: Vec<T>) -> u64 {
-            //println!("append entries!");
-            let mut buf: MessageBuf = MessageBuf::default();
             for entry in entries {
-                let _ = buf.push(AsBytes::as_bytes(&entry));
+                self.append_entry(entry);
             }
-            self.c_log.append(&mut buf).unwrap();
             self.get_log_len()
```
haraldng, line 220: This function can be done nicer I think.

Can use `let serialized = entries.into_iter().map()` to create a vec with all serialized entries. And then return `self.c_log.append(...) + 1` (if the +1 comment from `append_entry` is true)

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
-            let mut entries = vec![];
-            let zero_idx_to = to - from;
-            for (idx, msg) in buffer.iter().enumerate() {
-                if (idx as u64) >= zero_idx_to {
-                    break;
-                }
-                entries.push(FromBytes::read_from(msg.payload()).unwrap());
+            let buffer = match self.c_log.read(from, ReadLimit::default()) {
+                Ok(buf) => buf,
+                Err(_) => panic!(),
```
haraldng, line 233: use expect on all these and `unwrap()`

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
+                Ok(Some(prom_bytes)) => {
+                    let b_store: BallotStorage =
+                        FromBytes::read_from(prom_bytes.as_slice()).unwrap();
+                    Ballot {
+                        n: b_store.n,
+                        priority: b_store.priority,
+                        pid: b_store.pid,
+                    }
                 }
-                Ok(None) => Ballot::default(),
```
haraldng: Use this instead.

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
+                Err(_e) => panic!(),
             }
         }
 
         fn set_decided_idx(&mut self, ld: u64) {
-            let ld_bytes = AsBytes::as_bytes(&ld);
-            self.db.put(b"ld", ld_bytes).unwrap();
+            let ld_bytes = bincode::serialize(&ld).unwrap();
+            self.db.put(DECIDE, ld_bytes).unwrap();
         }
```
haraldng, line 289: Go back to the previous solution. We should use zerocopy when it is possible.

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
...
+                Ok(None) => 0,
+                Err(_e) => panic!(),
+            }
         }
 
-        fn get_stopsign(&self) -> Option<StopSignEntry> {
-            self.stopsign.clone()
+        fn set_compacted_idx(&mut self, trimmed_idx: u64) {
+            let trim_bytes = bincode::serialize(&trimmed_idx).unwrap();
+            self.db.put(TRIM, trim_bytes).unwrap();
```
haraldng, line 327: Use zerocopy here and when deserializing

omnipaxos_storage/src/memory.rs
```diff
@@ -1,239 +1,377 @@
 #[allow(missing_docs)]
 pub mod persistent_storage {
```
haraldng, line 2: This module is a bit misplaced, it is persistent but is under `memory`.

I would suggest either renaming this file to something more generic e.g. `storage.rs` or separating the memory and persistent storage modules into two different files.

haraldng (changes requested):
Most of the things I requested since last time are fixed, well done!

However, some of the changes to the file structure are unnecessary and confusing. Also please replace all `unwrap()` and empty `panic` with `expect` that includes some useful message. Furthermore, I think there might be a bug when returning the index from `append`. Double check that and see if it also is related to the trim problem.

omnipaxos_core/tests/reconfig_test.rs
```diff
@@ -1,31 +1,30 @@
...
-use util::TestSystem;
+use utils::util::{Value, SS_METADATA};
 
 #[cfg(feature = "hocon_config")]
 /// Verifies that the decided StopSign is correct and error is returned when trying to append after decided StopSign.
 #[test]
 #[serial]
 fn reconfig_test() {
-    let cfg = TestConfig::load("consensus_test").expect("Test config loaded");
+    let cfg = TestConfig::load("gc_test").expect("Test config loaded");
```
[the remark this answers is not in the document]
AbyelT: Though the configuration for consensus_test was only for `consensus_test.rs`, can revert it if this was intended.

omnipaxos_storage/src/memory.rs
AbyelT: `memory.rs` can be split into `persistent_storage.rs` and `memory_storage.rs`. PersistentStorage has grown large due to the wrapper structs used for storage and the new PersistentStorageConfig struct that were added.

haraldng: That is fine with me.

omnipaxos_storage/src/memory.rs
AbyelT: You are right, seems like the offsets in commitlog start from zero. Commitlog has a function `next_offset` that returns the next offset that is used in `get_log_len'. It can be used here but the above is much faster :)

omnipaxos_storage/src/memory.rs
AbyelT: Fortunately `MessageBuf`, the only type that `commitlog.append` accepts, lets you create it with an iterator :+1:

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+// Configuration for `PersistentStorage`.
+/// # Fields
+/// * `commitlog_path`: Path to the directory of the commitlog
+/// * `log_config`: Configuration of the commitlog
+/// * `rocksdb_path`: path to the rocksDB key-value store
+/// * `rocksdb_options` : Configuration of the rocksDB store
+pub struct PersistentStorageConfig {
+    commitlog_path: Option<String>,
+    commitlog_options: LogOptions,
+    rocksdb_path: Option<String>,
```
haraldng, line 85: Let's just use one `path` and use `COMMITLOG` and `ROCKSDB` under it.

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+        std::fs::metadata(&log_path)
+            .expect_err(&format!("commitlog already exists in {}", log_path));
+        std::fs::metadata(&rocks_path)
+            .expect_err(&format!("rocksDB store already exists in {}", rocks_path));
+
+        // Initialize Commitlog and rocksDB
+        let commitlog =
+            CommitLog::new(storage_config.commitlog_options).expect("Failed to create Commitlog");
+        let rocksdb = DB::open(&storage_config.rocksdb_options, &rocks_path)
+            .expect("Failed to create rocksDB store");
```
haraldng, line 187: For rocks I think you could use this [option](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_error_if_exists) and remove the check in the beginning. 

Furthermore, I'm not sure we need the check. Won't commitlog and rocks panic anyway if we try to create something on a path is already in use?

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+where
+    T: Entry + Serialize + for<'a> Deserialize<'a>,
+    S: Snapshot<T> + Serialize + for<'a> Deserialize<'a>,
+{
+    fn append_entry(&mut self, entry: T) -> u64 {
+        let entry_bytes = bincode::serialize(&entry).expect("Failed to serialize");
+        self.commitlog
+            .append_msg(entry_bytes)
+            .expect("Failed to append log entry")
+            + 1
```
haraldng, line 209:
```suggestion
            + 1  // +1 as commitlog returns the offset the entry was appended at, while we should return the index that the entry got in the log.
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn get_entries(&self, from: u64, to: u64) -> Vec<T> {
+        let buffer = self
+            .commitlog
+            .read(from, ReadLimit::default())
+            .expect("Failed to read from log");
+        let mut entries = Vec::<T>::with_capacity((to - from) as usize);
+        let mut iter = buffer.iter();
+        for _ in from..to {
+            let msg = iter.next().unwrap();
+            entries.push(bincode::deserialize(msg.payload()).expect("Failed to deserialize"));
```
haraldng, line 239: get_entries should either return either all entries between `from` and `to` if they exist in the log. If not, it should return an empty vec. Thus, this needs to be changed to something like this:
```suggestion
            match iter.next() {
                   Some(msg) => {
                        entries.push(bincode::deserialize(msg.payload()).expect("Failed to deserialize"));
                   },
                   None => {
                        return vec![];   // early return here
                   }     
            }
```

I also added a comment in the description for this function in the `Storage` trait to make it clear for others too.

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn get_log_len(&self) -> u64 {
+        self.commitlog.next_offset()
+    }
+
+    fn get_suffix(&self, from: u64) -> Vec<T> {
+        self.get_entries(from, self.commitlog.next_offset())
+    }
+
+    fn get_promise(&self) -> Ballot {
+        let value = self.rocksdb.get(NPROM).expect("Failed to retrive 'NPROM'");
```
haraldng, line 253:
```suggestion
        let promised = self.rocksdb.get(NPROM).expect("Failed to retrive 'NPROM'");
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn set_promise(&mut self, n_prom: Ballot) {
+        let ballot_store = BallotStorage::with(n_prom);
+        let prom_bytes = AsBytes::as_bytes(&ballot_store);
+        self.rocksdb
+            .put(NPROM, prom_bytes)
+            .expect("Failed to set 'NPROM'")
+    }
+
+    fn get_decided_idx(&self) -> u64 {
+        let value = self
```
haraldng, line 273:
```suggestion
        let decided = self
```

omnipaxos_core/src/storage.rs
```diff
@@ -110,13 +109,13 @@ where
     fn get_accepted_round(&self) -> Ballot;
 
     /// Returns the entries in the log in the index interval of [from, to)
```
haraldng, line 111:
```suggestion
    /// Returns the entries in the log in the index interval of [from, to). If entries **do not exist for the complete interval**, an empty Vector should be returned.
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+            .put(DECIDE, ld_bytes)
+            .expect("Failed to set 'DECIDE'");
+    }
+
+    fn get_accepted_round(&self) -> Ballot {
+        let value = self.rocksdb.get(ACC).expect("Failed to retrive 'ACC'");
+        match value {
+            Some(acc_bytes) => {
+                let b_store: BallotStorage =
+                    FromBytes::read_from(acc_bytes.as_slice()).expect("Failed to deserialize");
```
haraldng, line 297: Since the expect messages will be read by the user, it should be more clear what failed to deserialize and not just use the variable name (since they don't read the source code). **Please fix this for all expect messages in this file.**

```suggestion
                    FromBytes::read_from(acc_bytes.as_slice()).expect("Failed to deserialize the accepted ballot.");
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+        self.rocksdb
+            .put(STOPSIGN, stopsign)
+            .expect("Failed to set 'STOPSIGN'")
+    }
+
+    fn get_snapshot(&self) -> Option<S> {
+        let value = self
+            .rocksdb
+            .get(SNAPSHOT)
+            .expect("Failed to retive 'SNAPSHOT'");
```
haraldng, line 363:
```suggestion
            .expect("Failed to retrieve 'SNAPSHOT'");
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,384 @@
...
+        value.map(|snapshot_bytes| {
+            bincode::deserialize(snapshot_bytes.as_slice()).expect("Failed to deserialize")
+        })
+    }
+
+    fn set_snapshot(&mut self, snapshot: S) {
+        let stopsign = bincode::serialize(&snapshot).expect("Failed to set 'SNAPSHOT'");
```
haraldng, line 370:
```suggestion
        let stopsign = bincode::serialize(&snapshot).expect("Failed to serialize 'SNAPSHOT'");
```

docs/src/foreword.md
```diff
@@ -3,6 +3,6 @@ OmniPaxos is an in-development replicated log library implemented in Rust. OmniP
...
+The library consist of three parts: `omnipaxos_core`, `omnipaxos_runtime` and `omnipaxos_storage`. The `omnipaxos_core` implements the algorithms of OmniPaxos as plain Rust structs and is suitable for integration with systems that already have an async runtime or are implemented in an actor framework. It requires users to implement the interaction between the different structs themselves as we describe [here](ble/index.md). If you just want a replicated log out of the box, we suggest using `omnipaxos_runtime` instead, which hides all the interactions from the user by using [Tokio](https://tokio.rs/). Both core and runtime makes use of `omnipaxos_storage`, which offers storage implementations that can store the replicated state and log entries. More about storage is described in the 'Storage' section.
```
haraldng, line 6:
```suggestion
The library consist of three parts: `omnipaxos_core`, `omnipaxos_runtime` and `omnipaxos_storage`. The `omnipaxos_core` implements the algorithms of OmniPaxos as plain Rust structs and is suitable for integration with systems that already have an async runtime or are implemented in an actor framework. It requires users to implement the interaction between the different structs themselves as we describe [here](ble/index.md). If you just want a replicated log out of the box, we suggest using `omnipaxos_runtime` instead, which hides all the interactions from the user by using [Tokio](https://tokio.rs/). You can provide your own implementation for storing the log and state of OmniPaxos, but we also provide both in-memory and persistent storage implementations that work out of the box in `omnipaxos_storage`.
```

docs/src/sequencepaxos/storage.md
```diff
@@ -0,0 +1,208 @@
+# Storage
+You are free to use any storage implementation with `SequencePaxos`. The only requirement is that it implements the `Storage` trait. OmniPaxos includes the package `omnipaxos_storage` which provides two types of storage implementation, depending on the users need for fast reads and writes or persistance: `MemoryStorage` and `PersistentStorage`
```
haraldng, line 2:
```suggestion
You are free to use any storage implementation with `SequencePaxos`. The only requirement is that it implements the `Storage` trait. OmniPaxos includes the package `omnipaxos_storage` which provides two types of storage implementation that work out of the box: `MemoryStorage` and `PersistentStorage`.

If you do decide to implement your own storage, we recommend taking a look at `MemoryStorage` as a reference for implementing the functions required by `Storage`.
```

docs/src/sequencepaxos/storage.md
```diff
@@ -0,0 +1,208 @@
...
+        fn get_suffix(&self, from: u64) -> &[T] {
+            match self.log.get(from as usize..) {
+                Some(s) => s,
+                None => &[],
+            }
+        }
+
+        fn get_promise(&self) -> Ballot {
+            self.n_prom
+        }
```
haraldng, line 83: Please update this according to the changes you have made. For example `get_entries` now returns a vector instead of slice.

docs/src/sequencepaxos/storage.md
```diff
@@ -0,0 +1,208 @@
...
+        
+        ...
+    }
+```
+
+## PersistentStorage
+An persistent disk-based storage implementation, `PersistentStorage` makes use of the [Commitlog](https://crates.io/crates/commitlog) and [rocksDB](https://crates.io/crates/rocksdb) libraries to store logged entries and replica state. Users can configure the path to store log entries, replica state and storage related options through the `PersistentStorageConfig` struct.
```
haraldng, line 90:
```suggestion
`PersistentStorage` is a persistent storage implementation that uses [Commitlog](https://crates.io/crates/commitlog) to store the replicated log and [rocksDB](https://crates.io/crates/rocksdb) to store the state of OmniPaxos. Users can configure the path to store log entries, replica state, and storage related options through the `PersistentStorageConfig` struct. 
```

docs/src/sequencepaxos/storage.md
```diff
@@ -0,0 +1,208 @@
...
+        fn set_decided_idx(&mut self, ld: u64) {
+            let ld_bytes = AsBytes::as_bytes(&ld);
+            self.rocksdb
+                .put(DECIDE, ld_bytes)
+                .expect("Failed to set 'DECIDE'");
+        }
+            
+        ...
+    }
+```
```
haraldng, line 206: We don't need to show this to the user. We showed it for MemoryStorage just as a reference for them to understand how to implement the Storage trait. However, I would like you to show examples instead on how to use the constructor and set the different configurations for PersistentStorage.


```suggestion
```

haraldng (changes requested):
Very nice! I think the implementation looks good. There is one important comment on `get_entries` in PersistentStorage that needs to be fixed. Otherwise, it's mostly just tiny comments on the style and the documentation.

omnipaxos_storage/src/persistent_storage.rs
AbyelT: I agree the check is no longer needed, as both Commitlog and rocksDB will either create a new storage or open existing ones in a given path.

AbyelT:
Thanks for the feedback! so far I've managed to fix most issues and requested changes regarding persistent storage. What's left is to implement the new `recovery_test.rs` correctly and adding a new constructor to `persistentStorageConfig`.

haraldng (changes requested):

.gitignore
```diff
@@ -2,9 +2,7 @@
 /target
 /log
 Cargo.lock
-omnipaxos_core/rocksDB/
-omnipaxos_core/commitlog/
-clean_test.sh
+omnipaxos_core/storage/
```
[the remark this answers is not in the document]
AbyelT: I'll put the path so that `storage` is alongside its source code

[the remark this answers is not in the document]
haraldng: Perhaps not directly, next to it, but in a subfolder of its own.

omnipaxos_core/tests/recovery_test.rs
```diff
@@ -0,0 +1,325 @@
...
+        cfg.num_threads,
+        cfg.storage_type,
+        RECOVERY_PATH,
+    );
+
+    // check that the first proposals are decided
+    let mut vec_proposals = check_first_proposals(&sys, cfg.num_proposals, cfg.wait_timeout);
+
+    // kill and recover node
+    kill_and_recreate_node(&mut sys, &cfg, 3, RECOVERY_PATH);
```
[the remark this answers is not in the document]
AbyelT: Added a new function `check_leader_is_elected()` that checks if the leader is elected

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -6,14 +6,18 @@ use omnipaxos_core::{
...
-use std::{fs, iter::FromIterator, marker::PhantomData};
+use std::{iter::FromIterator, marker::PhantomData};
 use zerocopy::{AsBytes, FromBytes};
 
-const DEFAULT: &str = "/default_storage";
-const COMMITLOG: &str = "/commitlog";
-const ROCKSDB: &str = "/rocksDB";
+#[cfg(feature = "rocksdb")]
+use rocksdb::{Options, DB};
+#[cfg(not(feature = "rocksdb"))]
```
haraldng, line 15: Instead of using `#[cfg(not(feature = "rocksdb"))]` everywhere, use:
```suggestion
#[cfg(feature = "sled")]
```

And add this: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features somewhere so an error is thrown if somebody tries to use both sled and rocks at the same time.

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -102,76 +110,132 @@ impl PersistentStorageConfig {
...
             t: PhantomData::default(),
             s: PhantomData::default(),
         }
     }
+
+    // Creates a new storage instance, panics if a commitlog or rocksDB/sled instance exists in the given path
+    pub fn new(storage_config: PersistentStorageConfig) -> Self {
+        let path = storage_config
+            .path
+            .clone()
```
haraldng, line 225: I don't think you need to clone here:
```suggestion
            .as_ref()
```

omnipaxos_core/src/sequence_paxos.rs
```diff
@@ -1574,6 +1603,8 @@ impl ReconfigurationRequest {
     }
 }
 
+type PaxosAlias = (u64, u32);
```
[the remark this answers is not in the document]
AbyelT: My bad, i interpreted it as a type with `pid` and `configurationID` combined

omnipaxos_core/tests/recovery_test.rs
```diff
@@ -411,10 +376,12 @@ fn check_leader_is_elected(cfg: &TestConfig, leader_future: KFuture<Ballot>) ->
     (leader, follower)
 }
 
-// Propose and check that the first proposals before any node fails are decided
-// returns expected entries and a promise that the leader is elected
-fn check_first_proposals(sys: &TestSystem, cfg: &TestConfig) -> (Vec<Value>, KFuture<Ballot>) {
-    let (ble, px) = sys.ble_paxos_nodes().get(&1).unwrap();
+/// Propose and check that the first proposals before any node fails are decided
+fn check_initial_proposals(sys: &TestSystem, cfg: &TestConfig) -> Vec<Value> {
```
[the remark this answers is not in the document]
AbyelT: Good point, I'll let each test create the `proposals` vec and pass its reference down to any function that needs it since it will be used later e.g checking later proposals.

haraldng (approved):
Looks good to me.

omnipaxos_core/tests/config/test.conf
```diff
@@ -34,7 +33,6 @@ recovery_test {
     ble_hb_delay: 3
     increment_delay: 2
     num_proposals: 20
-    storage_type: persistent
```
haraldng: is this not needed anymore? I thought it would be used here:

https://github.com/haraldng/omnipaxos/blob/cfb37839a9e1abd04664ded86c3476f0b524e021/omnipaxos_core/tests/consensus_test.rs#L27

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,587 @@
...
+        for _ in from..to {
+            match iter.next() {
+                Some(msg) => {
+                    entries.push(
+                        bincode::deserialize(msg.payload())
+                            .expect("Failed to deserialize log entries"),
+                    );
+                }
+                None => {
+                    return vec![]; // early return here
```
Max-Meldrum, line 294: Is this early return here correct? Seems like it is filling things up into ``entries`` then once hitting None on the iterator, it returns an empty Vec instead of ``entries``

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,587 @@
...
+        self.get_entries(from, self.commitlog.next_offset())
+    }
+
+    fn get_promise(&self) -> Ballot {
+        #[cfg(feature = "rocksdb")]
+        {
+            let promised = self.rocksdb.get(NPROM).expect("Failed to retrieve 'NPROM'");
+            match promised {
+                Some(prom_bytes) => {
+                    let b_store: BallotStorage = FromBytes::read_from(prom_bytes.as_slice())
```
Max-Meldrum, line 315: It is also possible to use the zerocopy-derived struct directly as they are implemented for your types.
```rust
let b_store = BallotStorage::read_from(prom_bytes.as_slice())).expect(....);
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -0,0 +1,587 @@
...
+                    Ballot::with(b_store.n, b_store.priority, b_store.pid)
+                }
+                None => Ballot::default(),
+            }
+        }
+    }
+
+    fn set_promise(&mut self, n_prom: Ballot) {
+        let ballot_store = BallotStorage::with(n_prom);
+        let prom_bytes = AsBytes::as_bytes(&ballot_store);
```
Max-Meldrum, line 338:
```suggestion
        let prom_bytes = ballot_store.as_bytes();
```

omnipaxos_storage/src/persistent_storage.rs
haraldng: The function is correct according to the current API. The `get_entries()` is currently defined such that if the user tries to read from an index out of bounds, an empty vector is returned. This might be a bit counterintuitive so it might change in the future.

@AbyelT However, would it be possible first to check if all entries of the requested range really exist in the log before starting to read? So we don't start reading in the commitlog and then anyway end up returning `vec![]`.

docs/src/sequencepaxos/storage.md
```diff
@@ -0,0 +1,133 @@
...
+
+        fn get_promise(&self) -> Ballot {
+            self.n_prom
+        }
+        ...
+    }
+```
+
+## PersistentStorage
+`PersistentStorage` is a persistent storage implementation that stores the replicated log and the state of OmniPaxos. The struct uses [Commitlog](https://crates.io/crates/commitlog) to store replicated logs, for storing the OmniPaxos state users can choose between [rocksDB](https://crates.io/crates/rocksdb) or [sled](https://crates.io/crates/sled). Users can configure the path to log entries and OmniPaxos state, and storage-related options through `PersistentStorageConfig`. The configuraiton struct features a `default()` constructor for generating default configuration, and the constructor `with()` that takes path and storage options as arguments. We recommend the user to generate the default configuration and then setting the desired fields.
```
haraldng, line 94:
```suggestion
`PersistentStorage` is a persistent storage implementation that stores the replicated log and the state of OmniPaxos. The struct uses [Commitlog](https://crates.io/crates/commitlog) to store the replicated log, and the state is stored on [sled](https://crates.io/crates/sled) by default. The state can be changed to be stored on [RocksDB](https://crates.io/crates/rocksdb) instead of sled by using the feature`rocksdb`. Users can configure the path to log entries and OmniPaxos state, and storage-related options through `PersistentStorageConfig`. The configuration struct features a `default()` constructor for generating default configuration, and the constructor `with()` that takes path and storage options as arguments. 
```

docs/src/sequencepaxos/storage.md
```diff
@@ -0,0 +1,133 @@
...
+
+// create configuration with given arguments
+let my_config = PersistentStorageConfig::with(my_path, my_logopts, my_sled_opts);
+```
+## Selecting storage for OmniPaxos state
+`PersistentStorage` is configured to use `sled` by default, users can enable `rocksDB` as storage by adding `--features rocksdb` to their cargo command. The example below runs tests with rocksDB enabled:
+
+```no_run,noplaypen
+cargo test --features rocksdb
+```
```
haraldng, line 133: Show an example of how to import omnipaxos_storage with the rocksdb feature in `Cargo.toml` instead.

haraldng (changes requested):
Please fix changes in documentation and check comment on read_entries from Max.

omnipaxos_storage/src/persistent_storage.rs
AbyelT: Sure, the function can check if the endpoint `to` is greater than the current log length and thus out of range. That way we can move the early return to the beginning in the function.

docs/src/sequencepaxos/storage.md
```diff
@@ -91,7 +91,7 @@ An in-memory storage implementation, `MemoryStorage` offers fast reads and write
 ```
 
 ## PersistentStorage
```
haraldng, line 93: Could you add a subsection here that shows how to do fail-recovery? I.e. you re-intiate the persistent storage and recreate a SequencePaxos instance and then call `seq_paxos.fail_recovery()`

haraldng (changes requested):

docs/src/sequencepaxos/index.md
```diff
@@ -60,12 +60,22 @@ let raw_cfg = HoconLoader::new()
 let sp_config = SequencePaxosConfig::with_hocon(cfg);
 ```
 
-## Crash-recovery
-To support crash-recovery, we have to make sure that our storage implementation persisted the values. Then upon recovery, we have to make sure that our ``SequencePaxos`` will be started with the previously persisted state. To do so, create `SequencePaxos` as earlier described but use the persisted state as the `storage` argument. Then, call `fail_recovery()` to correctly initialize the volatile state.
+## Fail-recovery
+To support Fail-recovery, we have to make sure that our storage implementation can persist both the log entries and storage state. Upon recovery, we have to make sure that our ``SequencePaxos`` will start with the previously persisted state. To do so, we first re-create our storage with the same storage path as the previous instance. Then we create a `SequencePaxos` instance but use the persisted state as the `storage` argument. Lastly, we call `fail_recovery()` to correctly initialize the volatile state. We show an example using [`PersistentStorage`](storage.md), which gurantee persistency.
```
haraldng, line 64:
```suggestion
To support Fail-recovery, we must ensure that our storage implementation can persist both the log entries and storage state. Upon recovery, we have to make sure that our ``SequencePaxos`` will start with the previously persisted state. To do so, we first re-create our storage with the same storage path as the previous instance. Then we create a `SequencePaxos` instance but use the persisted state as the `storage` argument. Lastly, we call `fail_recovery()` to correctly initialize the volatile state. We show an example using [`PersistentStorage`](storage.md#persistentstorage).
```

docs/src/sequencepaxos/storage.md
```diff
@@ -126,11 +126,9 @@ my_sled_opts.new(true);
 let my_config = PersistentStorageConfig::with(my_path, my_logopts, my_sled_opts);
 ```
 ## Importing `omnipaxos_storage`
-To use the package, we need to add `omnipaxos_storage` to the dependencies in the cargo file. If a user wish to configure the options for the commit then the user must also add `commitlog` and either `rocksdb` or `sled` as dependencies. Otherwise the default constructor for `PersistentStorageConfig` lets the user create options for the commitlog and the state storage without importing. 
+To use the package, we need to add `omnipaxos_storage` to the dependencies in the cargo file. If a user wishes to configure the options for the commit log, then the user must add `commitlog` as a dependency. For configuring the state storage, either `rocksdb` or `sled` must also be added. Otherwise the default constructor for `PersistentStorageConfig` lets the user create default options for the commit log and the state storage without importing any dependencies.
```
haraldng, line 129: It makes more sense to show how to import omnipaxos_storage before showing MemoryStorage and PersistentStorage. So I suggest doing this modification and move it to here: https://github.com/haraldng/omnipaxos/blob/c4d3a83d46f8e78dac4e5cfe93e3eb226307f85e/docs/src/sequencepaxos/storage.md?plain=1#L3

```suggestion
To use the provided storage implementations, we need to add `omnipaxos_storage` to the dependencies in the cargo file.
```

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -96,6 +96,7 @@ pub struct PersistentStorageConfig {
     sled_options: Config,
 }
 
+#[allow(missing_docs)]
```
haraldng, line 99: Please add descriptions to public functions.

```suggestion
```

haraldng (changes requested):

docs/src/sequencepaxos/storage.md
```diff
@@ -1,6 +1,13 @@
 # Storage
 You are free to use any storage implementation with `SequencePaxos`. The only requirement is that it implements the `Storage` trait. OmniPaxos includes the package `omnipaxos_storage` which provides two types of storage implementation that work out of the box: `MemoryStorage` **and** `PersistentStorage`.
```
haraldng, line 2:
```suggestion
You are free to use any storage implementation with `SequencePaxos`. The only requirement is that it implements the `Storage` trait. OmniPaxos includes the package `omnipaxos_storage` which provides two types of storage implementation that work out of the box: `MemoryStorage` and `PersistentStorage`.
```

haraldng (approved):

haraldng (approved):

haraldng (approved):

omnipaxos_core/tests/recovery_test.rs
```diff
@@ -0,0 +1,408 @@
...
+        std::mem::take(&mut sys.kompact_system).expect("No KompactSystem in memory");
+    match kompact_system.shutdown() {
+        Ok(_) => {}
+        Err(e) => panic!("Error on kompact shutdown: {}", e),
+    };
+}
+
+#[test]
+#[serial]
+#[ignore = ""]
```
haraldng, line 68:
```suggestion
#[ignore]
```

omnipaxos_core/tests/recovery_test.rs
```diff
@@ -0,0 +1,408 @@
...
```
haraldng, line 170:
```suggestion
#[ignore]
```

omnipaxos_core/tests/recovery_test.rs
```diff
@@ -0,0 +1,408 @@
...
```
haraldng, line 117:
```suggestion
#[ignore]
```

omnipaxos_core/tests/recovery_test.rs
```diff
@@ -0,0 +1,408 @@
...
+use serial_test::serial;
+use std::{thread, time::Duration};
+use utils::{LatestValue, TestConfig, TestSystem, Value};
+
+const PERSISTENT_STORAGE: StorageTypeSelector = StorageTypeSelector::Persistent;
+const SLEEP_TIMEOUT: Duration = Duration::from_secs(1);
+
+#[test]
+#[serial]
+#[ignore = ""]
```
haraldng, line 15:
```suggestion
#[ignore]
```

haraldng (approved):

PR #53, opened 2022-09-01
Atomic storage & In-memory cache for state

### LINKED ISSUE #40: Internal in-memory storage for OmniPaxos state
Currently `OmniPaxos` reads directly from the user-provided `Storage` for variables such as the `promised_round` and `decided_idx`. This will be inefficient if `Storage` is storing these in persistent storage. Thus, `OmniPaxos` should have an internal storage struct that caches these values and reloads them from the persistent storage only upon failure recovery (might be related to #22)

AbyelT:

- [X] You ran `rustfmt` on the code base before submitting (on latest nightly with rustfmt support)
- [X] You reference which issue is being closed in the PR text (if applicable)

## Issues

Fix #40

## Breaking Changes

* SequencePaxos reads from in-memory cache struct for state
* Operations that write to storage are Atomic  

## Other Changes

* Added `CachedState` for internal in-memory state storage
* Struct implementing `Storage` trait now returns `Result` for all operations

### DIFF
--- a/omnipaxos_core/src/sequence_paxos.rs
+++ b/omnipaxos_core/src/sequence_paxos.rs
@@ -11,17 +11,11 @@ use super::{
 use crate::utils::hocon_kv::{CONFIG_ID, LOG_FILE_PATH, PEERS, PID, SP_BUFFER_SIZE};
 #[cfg(feature = "logging")]
 use crate::utils::logger::create_logger;
-use core::panic;
 #[cfg(feature = "hocon_config")]
 use hocon::Hocon;
 #[cfg(feature = "logging")]
 use slog::{debug, info, trace, warn, Logger};
 use std::{collections::Bound, fmt::Debug, marker::PhantomData, ops::RangeBounds, vec};
-
-/// Type alias for SequencePaxos Pid
-type Pid = u64;
-/// Type alias for SequencePaxos Configuration ID
-type ConfigurationID = u32;
 
 /// Type alias for SequencePaxos Pid
 type Pid = u64;
@@ -38,7 +32,7 @@ where
     B: Storage<T, S>,
 {
     storage: B,
-    cache: CachedState,
+    cached_state: CachedState,
     config_id: ConfigurationID,
     pid: Pid,
     peers: Vec<u64>, // excluding self pid
@@ -93,11 +87,11 @@ where
                 (state, 0, Ballot::default(), lds)
             }
         };
-        let cache = CachedState::default();
+        let cached_state = CachedState::default();
 
         let mut paxos = SequencePaxos {
             storage,
-            cache,
+            cached_state,
             config_id,
             pid,
             peers,
@@ -119,8 +113,8 @@ where
                 })
             },
         };
-        paxos.storage.set_promise(n_leader);
-        paxos.cache.set_promise(n_leader);
+        paxos.storage.set_promise(n_leader).expect("Falied to set promised ballot");
+        paxos.cached_state.set_promise(n_leader);
         #[cfg(feature = "logging")]
         {
             info!(paxos.logger, "Paxos component pid: {} created!", pid);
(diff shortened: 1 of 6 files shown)


omnipaxos_core/src/sequence_paxos.rs
```diff
@@ -220,16 +224,23 @@ where
...
         match c {
             Compaction::Trim(Some(idx)) if idx <= decided_idx && idx > compacted_idx => {
-                self.storage.trim(idx - compacted_idx);
-                self.storage.set_compacted_idx(idx);
+                self.storage.set_compacted_idx(idx).expect("Failed to set compacted index");
+                match self.storage.trim(idx - compacted_idx) {
+                    Ok(_) => self.cached_state.set_compacted_idx(idx),
+                    Err(_) => {
+                        self.storage.set_compacted_idx(self.cached_state.get_compacted_idx()).expect("Failed to restore compacted index");
+                        panic!();
```
haraldng, line 236: As I have pointed out before, it doesn't make sense to just panic or unwrap without any error message. Please add that.

omnipaxos_core/src/sequence_paxos.rs
```diff
@@ -612,16 +624,17 @@ where
...
-            let la = self.storage.get_log_len();
+            let na = self.cached_state.get_accepted_round();
+            let ld = self.cached_state.get_decided_idx();
+            let la = self.storage.get_log_len().expect("Failed to get log length");
             let my_promise = Promise::with(n, na, None, ld, la, self.get_stopsign());
             self.leader_state.set_promise(my_promise, self.pid);
             /* initialise longest chosen sequence and update state */
             self.state = (Role::Leader, Phase::Prepare);
-            let prep = Prepare::with(n, ld, self.storage.get_accepted_round(), la);
+            let prep = Prepare::with(n, ld, self.storage.get_accepted_round().expect("Failed to get accepted round"), la);
```
haraldng, line 637: Can't you just read from the cached state? Double check this for all gets.

```suggestion
            let prep = Prepare::with(n, ld, self.cached_state.get_accepted_round().expect("Failed to get accepted round"), la);
```

omnipaxos_core/src/sequence_paxos.rs
```diff
@@ -1344,25 +1414,40 @@ where
...
-            self.storage.set_decided_idx(self.storage.get_log_len() + 1);
+            let new_decided_idx = self.storage.get_log_len().expect("Failed to get log length") + 1;
+            self.storage.set_decided_idx(new_decided_idx).expect("Failed to set decided index");
+            match self.storage.set_stopsign(ss) {
+                Ok(_) => {
+                    self.cached_state.set_decided_idx(new_decided_idx)
+                },
+                Err(_) => {
+                    self.storage.set_decided_idx(self.cached_state.get_decided_idx()).expect("Failed to restore decided index");
+                },
```
haraldng, line 1439: shouldn't it panic here?

omnipaxos_core/src/sequence_paxos.rs
```diff
@@ -1276,11 +1335,20 @@ where
...
             if let Some(idx) = accsync.decide_idx {
-                self.storage.set_decided_idx(idx);
+                match self.storage.set_decided_idx(idx) {
+                    Ok(_) => {
+                        self.cached_state.set_accepted_round(accsync.n);
+                        self.cached_state.set_decided_idx(idx);
+                    },
+                    Err(_) => {
+                        self.storage.set_accepted_round(self.cached_state.get_accepted_round()).expect("Failed to restore accepted round");
+                    },
```
haraldng, line 1345: shouldn't it panic here? (and all other places where you have rolled-back)

haraldng (changes requested):
The main problem is the lack of consistency in the implementation. At some places, you read from the cache while at others you don't (when it is possible). Similar problem with when to panic after a rollback. So please double-check these things everywhere you have changed in the core.

Furthermore, I realized while reading the code that it would be nice to have the log length in the cached state.  Lastly, please make sure to load the values from `storage` to the cached state when doing `failure_recovery`.

haraldng:
Closing as the APi has changed (#57)

PR #58, opened 2023-03-15
Fix AccSync and Promise with Snapshots

JonathanArns:

## Issues
The issue that is addressed by this PR is an unaccounted for edge case in log synchronization when using snapshots.

Assume we are going to send an AccSync message.
The leader has a `decided_idx = 4` and a `compacted_idx = 4`, so his entire decided log is snapshotted and he does not have access to the individual entries anymore.
The follower has `decided_idx = 2`  and a `compacted_idx = 0`.

Currently the leader in this situation will try to create a delta snapshot from index 2 to 4 and a normal suffix for his further accepted entries.
The library will panic when trying to get the already compacted entries.
But even if we fix the panic, it is still not possible to create a correct delta snapshot, since we cannot access the required pieces of the log individually. Instead we should send a complete snapshot instead of the delta snapshot.

Analog problems also can occur when trying to construct a `Promise` message.

## Changes
To address this problem, this PR adds a few bounds checks or similar where necessary to prevent the mentioned panics.
Additionally, it adds the required logic to the leader and follower, to construct the correct `AccSync` and `Promise` messages when encountering the described edge case.

### Correctness
The described issue was discovered and the changes were tested using a DIY linearizability checker and the [ditm](https://github.com/JonathanArns/ditm) testing tool to simulate network partitions in a test cluster of 3 nodes.

### DIFF
--- a/omnipaxos_core/src/sequence_paxos/follower.rs
+++ b/omnipaxos_core/src/sequence_paxos/follower.rs
@@ -33,8 +33,18 @@ where
                     (None, suffix)
                 }
             } else if na == prep.n_accepted && accepted_idx > prep.accepted_idx {
-                let suffix = self.internal_storage.get_suffix(prep.accepted_idx);
-                (None, suffix)
+                if self.internal_storage.get_compacted_idx() > prep.accepted_idx
+                    && Self::use_snapshots()
+                {
+                    let delta_snapshot = self
+                        .internal_storage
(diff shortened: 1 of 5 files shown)


haraldng (approved):
Looks good to me.

JonathanArns:
I amended the commit without changes for the signature :)

PR #61, opened 2023-04-18
Feature auto reconnect

### LINKED ISSUE #56: Id for Accept messages
Enumerate `Accept` messages with an `id: u64`. This enables us to discover and react to a dropped message rather than relying on the user to call the `reconnected()`.

The second part is to handle follower-follower reconnections which must be handled because during the disconnection one of them might have become the leader, and therefore we need to do the preparereq when reconnecting. This can be done through the BLE when we receive a ballot from a node that we previously did not.

kevin-harrison:

## Issues
fix #56
Addresses leader-to-follower reconnections using accept message sequence numbers. The issue of follower-to-follower reconnections remains open.

## Other Changes
Changed `OmniPaxosComponent` for Kompact testing
 - Decided futures now only consumes decided entries made after the future was issued
 - Decided futures will receive dummy values for values trimmed due to deciding a snaphot
 - Added support for simulating network faults to other Components with `set_connection()`

### DIFF
--- a/omnipaxos_core/src/messages.rs
+++ b/omnipaxos_core/src/messages.rs
@@ -9,7 +9,7 @@ pub mod sequence_paxos {
     use crate::{
         ballot_leader_election::Ballot,
         storage::{Entry, Snapshot, SnapshotType, StopSign},
-        util::NodeId,
+        util::{NodeId, SequenceNumber},
     };
     use std::fmt::Debug;
 
@@ -59,7 +59,7 @@ pub mod sequence_paxos {
         /// The current round.
         pub n: Ballot,
         /// The sequence number of this message in the leader-to-follower accept sequence
-        pub seq_num: u64,
+        pub seq_num: SequenceNumber,
         /// The decided snapshot.
         pub decided_snapshot: Option<SnapshotType<T, S>>,
         /// The log suffix.
@@ -81,7 +81,7 @@ pub mod sequence_paxos {
         /// The current round.
         pub n: Ballot,
         /// The sequence number of this message in the leader-to-follower accept sequence
-        pub seq_num: u64,
+        pub seq_num: SequenceNumber,
         /// The decided index.
         pub decided_idx: u64,
         /// Entries to be replicated.
@@ -103,7 +103,7 @@ pub mod sequence_paxos {
         /// The current round.
         pub n: Ballot,
         /// The sequence number of this message in the leader-to-follower accept sequence
-        pub seq_num: u64,
+        pub seq_num: SequenceNumber,
         /// The decided index.
         pub decided_idx: u64,
     }
--- a/omnipaxos_core/src/messages.rs
+++ b/omnipaxos_core/src/messages.rs
@@ -58,6 +58,8 @@ pub mod sequence_paxos {
     {
         /// The current round.
         pub n: Ballot,
+        /// The sequence number of this message in the leader-to-follower accept sequence
+        pub seq_num: u64,
         /// The decided snapshot.
         pub decided_snapshot: Option<SnapshotType<T, S>>,
         /// The log suffix.
@@ -85,6 +87,8 @@ pub mod sequence_paxos {
     {
         /// The current round.
         pub n: Ballot,
+        /// The sequence number of this message in the leader-to-follower accept sequence
+        pub seq_num: u64,
         /// The decided index.
         pub decided_idx: u64,
         /// Entries to be replicated.
--- a/omnipaxos_core/src/messages.rs
+++ b/omnipaxos_core/src/messages.rs
@@ -102,6 +102,8 @@ pub mod sequence_paxos {
     pub struct Decide {
         /// The current round.
         pub n: Ballot,
+        /// The sequence number of this message in the leader-to-follower accept sequence
+        pub seq_num: u64,
         /// The decided index.
         pub decided_idx: u64,
     }
--- a/omnipaxos_core/src/messages.rs
+++ b/omnipaxos_core/src/messages.rs
@@ -72,13 +72,6 @@ pub mod sequence_paxos {
         pub stopsign: Option<StopSign>,
     }
 
-    /// The first accept message sent. Only used by a pre-elected leader after reconfiguration.
-    #[derive(Clone, Debug)]
-    pub struct FirstAccept {
-        /// The current round.
-        pub n: Ballot,
-    }
-
     /// Message with entries to be replicated and the latest decided index sent by the leader in the accept phase.
     #[derive(Clone, Debug)]
     pub struct AcceptDecide<T>
@@ -158,7 +151,6 @@ pub mod sequence_paxos {
         Prepare(Prepare),
         Promise(Promise<T, S>),
         AcceptSync(AcceptSync<T, S>),
-        FirstAccept(FirstAccept),
         AcceptDecide(AcceptDecide<T>),
         Accepted(Accepted),
         Decide(Decide),
--- a/omnipaxos_core/src/sequence_paxos/follower.rs
+++ b/omnipaxos_core/src/sequence_paxos/follower.rs
@@ -33,8 +33,18 @@ where
                     (None, suffix)
                 }
             } else if na == prep.n_accepted && accepted_idx > prep.accepted_idx {
-                let suffix = self.internal_storage.get_suffix(prep.accepted_idx);
-                (None, suffix)
+                if self.internal_storage.get_compacted_idx() > prep.accepted_idx
+                    && Self::use_snapshots()
+                {
+                    let delta_snapshot = self
+                        .internal_storage
+                        .create_diff_snapshot(prep.decided_idx, decided_idx);
+                    let suffix = self.internal_storage.get_suffix(decided_idx);
+                    (Some(delta_snapshot), suffix)
+                } else {
+                    let suffix = self.internal_storage.get_suffix(prep.accepted_idx);
+                    (None, suffix)
+                }
             } else {
                 (None, vec![])
             };
--- a/omnipaxos_core/src/sequence_paxos/follower.rs
+++ b/omnipaxos_core/src/sequence_paxos/follower.rs
@@ -187,6 +187,12 @@ where
 
     pub(crate) fn handle_decide(&mut self, dec: Decide) {
         if self.internal_storage.get_promise() == dec.n && self.state.1 == Phase::Accept {
+            if self.current_seq_num + 1 != dec.seq_num {
+                // If accept sequence is broken reconnect to leader instead
+                self.reconnected(dec.n.pid);
+                return;
+            }
+            self.current_seq_num = dec.seq_num;
             self.internal_storage.set_decided_idx(dec.decided_idx);
         }
     }
--- a/omnipaxos_core/src/sequence_paxos/follower.rs
+++ b/omnipaxos_core/src/sequence_paxos/follower.rs
@@ -2,10 +2,9 @@ use super::super::ballot_leader_election::Ballot;
 
 use super::*;
 
-use crate::storage::SnapshotType;
-use crate::util::MessageStatus;
+use crate::{storage::SnapshotType, util::MessageStatus};
 #[cfg(feature = "logging")]
-use slog::debug;
+use slog::warn;
 
 impl<T, S, B> SequencePaxos<T, S, B>
 where
@@ -192,7 +191,10 @@ where
             match msg_status {
                 MessageStatus::First => {
                     #[cfg(feature = "logging")]
-                    warn!("Decide cannot be the first message in a sequence!");
+                    warn!(
+                        self.logger,
+                        "Decide cannot be the first message in a sequence!"
+                    );
                     return;
                 }
                 MessageStatus::Expected => self.current_seq_num = dec.seq_num,
--- a/omnipaxos_core/src/sequence_paxos/follower.rs
+++ b/omnipaxos_core/src/sequence_paxos/follower.rs
@@ -142,26 +142,23 @@ where
         }
     }
 
-    pub(crate) fn handle_firstaccept(&mut self, f: FirstAccept) {
-        #[cfg(feature = "logging")]
-        debug!(self.logger, "Incoming message First Accept");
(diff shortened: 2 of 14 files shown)


omnipaxos_core/src/util.rs
```diff
@@ -45,6 +45,8 @@ where
 {
     pub n_leader: Ballot,
     pub promises_meta: Vec<Option<PromiseMetaData>>,
+    // the sequence number of accepts for each follower where AcceptSync has sequence number = 1
+    pub accept_sequences: Vec<u64>,
```
haraldng, line 49: Same here with naming:

```suggestion
    pub follower_seq_nums: Vec<u64>,
```

omnipaxos_core/src/sequence_paxos/follower.rs
```diff
@@ -100,6 +101,12 @@ where
             self.internal_storage.set_accepted_round(accsync.n);
             self.internal_storage.set_decided_idx(accsync.decided_idx);
             self.state = (Role::Follower, Phase::Accept);
+            // Begin accept sequence
+            if self.accept_sequence != 0 {
+                panic!("AcceptSync must be first message in accept sequence!");
+            }
```
haraldng, line 107: Since we have passed the check that we are in the prepare phase (where we set the seq number to 0), this can be removed.

```suggestion
```

omnipaxos_core/src/sequence_paxos/follower.rs
```diff
@@ -156,6 +163,15 @@ where
         if self.internal_storage.get_promise() == acc.n
             && self.state == (Role::Follower, Phase::Accept)
         {
+            // If accept sequence is broken reconnect to leader instead
+            if self.accept_sequence == 0 {
+                panic!("AcceptDecide cannot be first message in accept sequence!");
```
haraldng, line 168:
```suggestion
```

omnipaxos_core/src/util.rs
```diff
@@ -88,6 +91,27 @@ where
...
+        self.accept_sequences[Self::pid_to_idx(pid)] = 0;
+    }
+
+    pub fn get_acceptsync_sequence_num(&mut self, pid: NodeId) -> u64 {
+        if self.accept_sequences[Self::pid_to_idx(pid)] != 0 {
+            panic!("AcceptSync must be first message in accept sequence");
+        }
+        self.accept_sequences[Self::pid_to_idx(pid)] = 1;
+        1
+    }
```
haraldng, line 105: This function is confusing since it is called `get_`... but actually also sets the sequence number of the node on line 103. I also think the function is unnecessary, and it would be more clean and clear to just use a constant for the AcceptSync sequence number (and set/reset) where needed.

`const ACCSYNC_SEQNUM: u64 = 1;`

```suggestion
```

omnipaxos_core/src/util.rs
```diff
@@ -88,6 +91,27 @@ where
...
+
+    pub fn next_acceptdecide_sequence_num(&mut self, pid: NodeId) -> u64 {
+        if self.accept_sequences[Self::pid_to_idx(pid)] == 0 {
+            panic!("AcceptDecide cannot be first in accept sequence.");
+        }
```
haraldng, line 110:
```suggestion
    pub fn next_acceptdecide_sequence_num(&mut self, pid: NodeId) -> u64 {
```

omnipaxos_core/tests/reconnect_test.rs
```diff
@@ -0,0 +1,260 @@
...
+use std::{thread, time::Duration};
+use utils::{LatestValue, TestConfig, TestSystem, Value};
+
+const SLEEP_TIMEOUT: Duration = Duration::from_secs(1);
+
+/// Verifies that a leader sends out AcceptSync messages
+/// with increasing sequence numbers.
+#[test]
+#[serial]
+fn ascending_accept_sequence_test() {
```
haraldng, line 20:
```suggestion
fn increasing_accept_sequence_test() {
```

omnipaxos_core/tests/utils.rs
```diff
@@ -402,6 +402,19 @@ impl TestSystem {
...
+        let node = self
+            .nodes
+            .get(&pid)
+            .expect(&format!("Cannot find node {pid}"));
+        self.kompact_system
+            .as_ref()
+            .expect("No KompactSystem found!")
+            .stop_notify(node)
+            .wait_timeout(STOP_COMPONENT_TIMEOUT)
+            .expect("ReplicaComp never started!");
```
haraldng, line 415:
```suggestion
            .expect("ReplicaComp never stopped!");
```

omnipaxos_core/tests/utils.rs
```diff
@@ -544,12 +573,21 @@ pub mod omnireplica {
...
-                                .unwrap()
-                                .reply(s.snapshot.value)
-                                .expect("Failed to reply promise!"),
+                            LogEntry::Snapshotted(s) => {
+                                // Reply with dummy value for futures which were trimmed away
+                                for _ in 1..(s.trimmed_idx - self.decided_idx) {
+                                    self.decided_futures
+                                        .pop()
+                                        .unwrap()
+                                        .reply(Value(0))
```
haraldng, line 582: Use a constant `SNAPSHOTTED_DECIDE = Value(0)` instead.

omnipaxos_core/tests/reconnect_test.rs
```diff
@@ -0,0 +1,260 @@
...
+    let mut sys = TestSystem::with(
+        cfg.num_nodes,
+        cfg.election_timeout,
+        cfg.num_threads,
+        cfg.storage_type,
+    );
+    sys.start_all_nodes();
+
+    // Propose some values so that a leader is elected
+    make_proposals(&sys, &cfg, 1, 1, 5);
```
haraldng, line 32: Make all these kind of test variables into a variable/constant defined in the beginning of the test so it's easy to read the code and get an idea of what will happen in the test

omnipaxos_core/tests/reconnect_test.rs
```diff
@@ -0,0 +1,260 @@
...
+            x.decided_futures.push(Ask::new(kprom, ()));
+        });
+        proposal_futures.push(kfuture);
+    }
+
+    match FutureCollection::collect_with_timeout::<Vec<_>>(proposal_futures, cfg.wait_timeout) {
+        Ok(_) => {}
+        Err(e) => panic!("Error on collecting futures of decided proposals: {}", e),
+    }
+}
```
haraldng, line 176: I'm not sure if we already have something similar in `tests/util.rs`. If we have use them, else move these there so we can reuse them in other tests.

omnipaxos_core/tests/reconnect_test.rs
```diff
@@ -0,0 +1,260 @@
...
+            e => panic!(
+                "{}",
+                format!(
+                    "Unexpected entry at idx {}: {:?}, decided_idx: {}",
+                    idx, e, decided_idx
+                )
+            ),
+        }
+    }
+}
```
haraldng, line 260: I'm pretty sure we have similar functions either in util or consensus_test. Please reuse them or modify them so you don't have to rewrite them here.

haraldng (changes requested):
Nice work! Most of it looks good, but there is still some stuff to do:
1. Consistency in naming: all code related to the sequence numbers should be consistently named in the messages and state in the leader and follower. I suggest we base it on `seq_num`, but use prefixes and suffixes when needed e.g., `follower_seq_nums` in the leader state.
2. With the sequence numbers we can now remove `FirstAccept`. Instead initialise the servers to be in the Accept phase and set the seq_nums accordingly: https://github.com/haraldng/omnipaxos/blob/5bf1454735cb77144c857b199b2e9bbdb822cfb0/omnipaxos_core/src/sequence_paxos/mod.rs#L78
3. Naming and refactoring in tests. Use constants and variable names so that it becomes easy to understand the test cases just by reading the code. Furthermore, some of the helper functions for checks either already exist or are very similar to the ones that exist. So please reuse them to keep the code base minimal and clean.

omnipaxos_core/src/util.rs
kevin-harrison: I think the LeaderState API just having `reset_seq_num` and `next_seq_num` could be clean. Then when sending accepts you just ask for the next seq number and when sending AcceptSyncs you reset the sequence and then ask for the next seq number. It becomes less clear that AcceptSync always has seq 1, but you don't have to define a third function `set_seq_num` and don't have users of LeaderState need to know what value to set the sequence to.

omnipaxos_core/src/messages.rs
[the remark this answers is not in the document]
kevin-harrison: Is the idea to send an `AcceptDecide` with `seq_num` = 1 as a psuedo-`AcceptSync` instead of sending a `FirstAccept`? 

Also `FirstAccept`s cause followers to set their accepted round based on the message. What is the purpose of this? I'm confused about what accepted round is for compared to promises and ballot numbers.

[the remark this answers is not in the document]
haraldng: Yes, replace FirstAccept with an AcceptDecide with seq_num=1.

The accepted ballot needs to be set to know in which round the server last appended to the log, which is needed in the prepare phase so that the new leader can see which server is most updated and adopt that log.

haraldng (changes requested):
Looks mostly good. We need to add `seq_num` to the `Decide` message.

omnipaxos_core/src/sequence_paxos/follower.rs
```diff
@@ -147,30 +142,25 @@ where
...
+            if acc.seq_num == 1 {
+                // psuedo-AcceptSync for reconfigurations
+                #[cfg(feature = "logging")]
+                debug!(
+                    self.logger,
+                    "Incoming message first Accept of reconfiguration"
+                );
+                self.internal_storage.set_accepted_round(acc.n);
+                self.forward_pending_proposals();
+            } else if self.current_seq_num + 1 != acc.seq_num {
```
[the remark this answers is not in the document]
kevin-harrison: Even if the seq_num != 1 we still need to check for sequence breaks with the else if

haraldng (changes requested):
Fix small comments and make seq_num be a tuple that also includes a "session number" that we increment after every sync.

omnipaxos_core/src/sequence_paxos/follower.rs
```diff
@@ -148,16 +148,11 @@ where
         if self.internal_storage.get_promise() == acc.n
             && self.state == (Role::Follower, Phase::Accept)
         {
-            if acc.seq_num == ACCSYNC_SEQ_NUM {
+            if acc.seq_num.counter == PREDEFINED_LEADER_FIRST_ACCEPT {
```
haraldng, line 151: I guess this is always (0,1) so we can change `PREDEFINED_LEADER_FIRST_ACCEPT` to that.

omnipaxos_core/src/sequence_paxos/follower.rs
```diff
@@ -148,16 +148,11 @@ where
...
                 // psuedo-AcceptSync for reconfigurations
-                #[cfg(feature = "logging")]
-                debug!(
-                    self.logger,
-                    "Incoming message first Accept of reconfiguration"
-                );
                 self.internal_storage.set_accepted_round(acc.n);
                 self.forward_pending_proposals();
-            } else if self.current_seq_num + 1 != acc.seq_num {
+            } else if !acc.seq_num.is_successor_of(self.current_seq_num) {
```
haraldng, line 155: To avoid doing `reconnected()` due to old messages with lower session or seq numbers I suggest we have a function:

```
enum MessageStatus {
      Expected,   // happy case
      DroppedPreceding,    // do reconnected
      Outdated,   // lower SequenceNumber can be ignored
}

fn check_msg_status(&mut self, msg_seq_num: SequenceNumber) -> MessageStatus 
```

haraldng (changes requested):

omnipaxos_core/src/util.rs
```diff
@@ -301,3 +301,23 @@ pub type TrimmedIndex = u64;
 pub type NodeId = u64;
 /// ID for an OmniPaxos configuration (i.e., the set of servers in an OmniPaxos cluster)
 pub type ConfigurationId = u32;
+
+#[allow(missing_docs)]
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
+pub struct SequenceNumber {
+    pub round: u64,
```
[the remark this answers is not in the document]
kevin-harrison: Yeah I agree round is probably not the correct name. My thinking in changing it away from session is that its value can increase even without a new TCP session. For example leader loses leadership then regains leadership and syncs again. Not sure what to call it, `accept_seq_round`?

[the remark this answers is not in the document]
kevin-harrison: On second thought maybe session works since in the context of that specific ballot it does represent TCP sessions.

omnipaxos_core/src/sequence_paxos/mod.rs
```diff
@@ -346,7 +346,7 @@ where
             return;
         } else if pid == self.leader.pid {
             self.state = (Role::Follower, Phase::Recover);
-            self.current_seq_num = 0;
+            self.current_seq_num.next_round();
```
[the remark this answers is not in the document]
kevin-harrison: How about we keep this and use the fact that we have incremented the `round` (`session`) to allow the follower to avoid sending multiple `PrepareReq`. So if the follower gets (1,2) (1,4) (1,5) only receiving (1,4) it will go to (2,0) and then (1,5) can be ignored.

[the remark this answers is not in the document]
haraldng: I think (1,5) will be ignored anyway in `handle_acceptdecide` because it is in the `Recover` state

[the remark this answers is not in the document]
haraldng: The point is also that I think it is better if only the leader sets the seq numbers and followers don't touch it and only check if it got the expected message.

haraldng (changes requested):
Add some docs and remove panic and this should be good to go.

haraldng (approved):

PR #80, opened 2023-04-27
WIP: demo with network actor

JonathanArns:
The demo currently builds and runs, but does not do anything besides sending lots of heartbeats and electing a leader at the start.
The goal is to be able to demonstrate behavior under partial connectivity, possibly compared with a raft implementation of the same demo.
To visualize, a small dashboard that shows information like throughput and network connectivity status would be nice.
Ideally we would like an interactive client as well as a sort of test-bed client, to possibly replace Kompics for tests with this eventually.

### TODO
- [ ] partial connectivity
- [ ] interactive client
- [ ] dashboard to show throughput, partitions, ...
- [ ] test-bed client
- [ ] Raft comparison implementation

- [ ] You ran the local CI checker with `./check.sh` with no errors
- [ ] You reference which issue is being closed in the PR text (if applicable)


haraldng:
For Raft, here's some examples of how to use it: https://github.com/haraldng/kompicsbenches/blob/duration-based/kompact/src/bench/atomic_broadcast/raft.rs#L630

haraldng:
For the dashboard, we previously considered `tui`: https://crates.io/crates/tui.
I think tui might not be actively maintained, but there is a new fork of it which is currently active, so maybe we should use it instead: https://crates.io/crates/ratatui

I also think it's a good idea if @TimXLHan can help look into the dashboard part.

haraldng:
I think we said that @kevin-harrison was going to look into adding RocksDB to the demo, but I had some time during the weekend and added it. Changes include:
- Added RocksDB wrapped in a struct `Database`
- Changed the entry type in OmniPaxos from `KeyValue` to `KVCommand` so that we can do `put` and `delete` in the demo.
- Added handler in `server` for applying changes from the omnipaxos log to rocksdb.

haraldng:
Closing this for now as we continue in a dedicated repo.

PR #82, opened 2023-04-28
Make Snapshot an associated type in Entry

### LINKED ISSUE #74: Documentation for features
Add a page that describes all the different feature flags that can be set.

haraldng:

## Issues
Remove awkward Snapshot generic type and fix #74 

This PR aims to improve usability by getting rid of one of the many generic types we have. This makes writing Storage and OmniPaxos a lot less verbose, i.e.,
`OmniPaxos<Entry, MemoryStorage<Entry>>` instead of `OmniPaxos<Entry, Snapshot, MemoryStorage<Entry, Snapshot>>`
## Breaking Changes
- OmniPaxos and all structs that had `<T: Entry, S: Snapshot>` as generic types are now replaced with `<T: Entry>`. Instead the `Entry` trait now has an associated type `Storage`. So all usages of `S` in the  API has been replaced  with `T::Snapshot`.
- Removed the blanket implementation for `Entry` for all types that implemented `Clone + Debug`. To implement Entry, the user can derive it using `omnipaxos_macros::Entry`, or implement it manually if they want to use snapshots.

## Other Changes
- Updated documentation, examples, and all-things related to this API change. Also added documentation for feature flags.
- Added the workspace `omnipaxos_macros` and tests for it.

### DIFF
--- a/examples/kv_store/Cargo.toml
+++ b/examples/kv_store/Cargo.toml
@@ -6,11 +6,10 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-omnipaxos_core = { path = "../../omnipaxos_core" }
+omnipaxos_core = { path = "../../omnipaxos_core", features = ["macros"] }
 omnipaxos_storage = { path = "../../omnipaxos_storage" }
-omnipaxos_macros = { path = "../../omnipaxos_macros", optional = true }
 tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros"] }
 serde = { version = "1.0", features = ["derive"] }
 
 [features]
-derive_entry = ["omnipaxos_macros"]
+derive_entry = []
--- a/examples/kv_store/Cargo.toml
+++ b/examples/kv_store/Cargo.toml
@@ -8,5 +8,9 @@ edition = "2021"
 [dependencies]
 omnipaxos_core = { path = "../../omnipaxos_core" }
 omnipaxos_storage = { path = "../../omnipaxos_storage" }
+omnipaxos_macros = { path = "../../omnipaxos_macros", optional = true }
 tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros"] }
 serde = { version = "1.0", features = ["derive"] }
+
+[features]
+derive_entry = ["omnipaxos_macros"]
--- a/omnipaxos_core/src/storage.rs
+++ b/omnipaxos_core/src/storage.rs
@@ -13,8 +13,13 @@ use std::{
 
 /// Type of the entries stored in the log.
 pub trait Entry: Clone + Debug {
+    #[cfg(not(feature = "serde"))]
     /// The snapshot type for this entry type.
     type Snapshot: Snapshot<Self>;
+
+    #[cfg(feature = "serde")]
+    /// The snapshot type for this entry type.
+    type Snapshot: Snapshot<Self> + Serialize + for<'a> Deserialize<'a>;
 }
 
 /// A StopSign entry that marks the end of a configuration. Used for reconfiguration.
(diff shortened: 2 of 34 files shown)


Max-Meldrum:
To me, it seems the ``Snapshot`` type is defined at the wrong level after looking at the kv example. However, correct me if I am wrong. The impression I get is that a ``Entry`` defines the type the log is working with. Whereas the ``Snapshot`` trait should tell the ``Storage`` or ``SequencePaxos`` layer how to take a snapshot of possibly many entries ( ``&[impl Entry]``). Right now, the snapshotting is defined on the entry type and not at the a higher level.

Possible options:
1. Move it to the Storage trait instead

```rust
pub trait Storage {
   type Snapshot;
}
```
2.  Let users define snapshotting at the ``OmniPaxos`` struct level.

```rust
pub struct OmniPaxos<T, B, S = NoSnapshot> {}

impl<T, B, NoSnapshot> OmniPaxos<T, B, NoSnapshot> {
}

impl<T, B, S> OmniPaxos<T, B, S> {
    pub const fn with_snapshot(snapshot: S) -> OmniPaxos<T, B, S> {
        Self { ... }
    }
}

```

haraldng:
Snapshots are not necessarily only related to how data is stored, it also relates to how data is transferred in the protocol. For instance, a few of the [messages](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos_core/src/messages.rs#L47) have fields with the snapshot, and they are agnostic of the storage (i.e., don't have Storage as generic type).

I have also thought of having different implementations of OmniPaxos depending on if Snapshot is used or not. But since we have so many message handlers (split into 2-3 files `sequence_paxos/{mod.rs, leader.rs, follower.rs}`), this would either require a lot of duplicate code or big surgery on the existing code so we can share functions between the implementations. That would risk correctness and maintainability so I don't think it's worth it. But do let me know if there is an easy approach!

omnipaxos_core/src/storage.rs
```diff
@@ -63,18 +70,17 @@ impl PartialEq for StopSign {
...
+pub enum SnapshotType<T>
 where
     T: Entry,
-    S: Snapshot<T>,
 {
-    Complete(S),
-    Delta(S),
+    Complete(T::Snapshot),
+    Delta(T::Snapshot),
     _Phantom(PhantomData<T>),
```
Max-Meldrum, line 79: Is this Phantom variant still needed?

omnipaxos_core/src/storage.rs
```diff
@@ -63,18 +70,17 @@ impl PartialEq for StopSign {
...
 }
 
 /// Functions required by Sequence Paxos to implement snapshot operations for `T`. If snapshot is not desired to be used, use the unit type `()` as the Snapshot parameter in `SequencePaxos`.
```
Max-Meldrum, line 82: Outdated comment?

examples/kv_store/Cargo.toml
```diff
@@ -8,5 +8,9 @@ edition = "2021"
 [dependencies]
 omnipaxos_core = { path = "../../omnipaxos_core" }
 omnipaxos_storage = { path = "../../omnipaxos_storage" }
+omnipaxos_macros = { path = "../../omnipaxos_macros", optional = true }
```
Max-Meldrum, line 11: It is common to re-export the macros from within the main crate.

```toml
# omnipaxos_core/Cargo.toml
[features]
# derive/macros
derive = ["omnipaxos_macros"]

[dependencies]
omnipaxos_macros = { path = "../", optional = true }
```

```rust
// omnipaxos_core/lib.rs
// example from serde: https://github.com/serde-rs/serde/blob/v1.0.89/serde/src/lib.rs#L245-L256
#[cfg(feature = "derive")]
#[allow(unused_imports)]
#[macro_use]
extern crate omnipaxos_macros;
#[cfg(feature = "derive")]
#[doc(hidden)]
pub use omnipaxos_macros::*;
```

```toml
[dependencies]
# if not enabled by default
omnipaxos_core = { ... , features = ["derive"] }
```

haraldng: I found re-exporting in the root a bit confusing as the macro would be `omnipaxos_core::Entry` and the actual trait would be `omnipaxos_core::storage::Entry`. 
So I decided to re-export it in a new mod: `omnipaxos_core::macros::Entry`.

haraldng:
Thanks a lot for the comments @Max-Meldrum, really appreciate it!

Issue #59, opened 2023-04-08
Batching in Storage


haraldng:
Wait until X number of entries before doing `append_entries` in Storage and then reply with Accepted. Might need to reintroduce `tick()` so that we can perform the append after a timeout without getting X entries.

PR #84, opened 2023-04-28
Batching

TimXLHan:

## Issues

This PR fix https://github.com/haraldng/omnipaxos/issues/40 and fix https://github.com/haraldng/omnipaxos/issues/59.

## Breaking Changes

- **States cache implementation**

  There is an implementation of abstraction of memory cache called `StateCache` to store simple state values of OmniPaxos and batched log entries.

- **Batch processing implementation**

  Changed the return type of `InternalStorage::append_entries` from `u64` ot `Option<u64>`. Means when size of batched logs reaches the `batch_size` batched logs will be flushed into storage and the function will return actual log length, otherwise return `None`.

- **Test for batching**

  See `omnipaxos_core/tests/batching_test.rs`

## Other Changes

Added a config called `batch_size` with default value of `1` into `OmniPaxosConfig` and updated all the related codes.


omnipaxos_core/src/storage.rs
```diff
@@ -355,52 +442,83 @@ where
...
+        self.state_cache.append_entry(entry);
+        if self.state_cache.get_batching_size() >= self.batch_size as u64 {
+            self.flush_batch();
+        }
+        self.get_log_len()
+    }
+
+    // Append entries in batch, if the batch size is reached, flush the batch and return the
+    // accepted index
+    pub(crate) fn append_entries(&mut self, entries: Vec<T>) -> Option<u64> {
```
haraldng, line 465: There are some cases where we want to actually append everything directly without batching (i.e., when synchronizing in the Prepare phase), so I suggest we add a parameter `skip_batching`:
```suggestion
    pub(crate) fn append_entries(&mut self, entries: Vec<T>, skip_batching: bool) -> Option<u64> {  
     // flush directly if skip_batching is true
```

omnipaxos_core/src/sequence_paxos/leader.rs
haraldng, line 1: We should use the `accepted_index` that we get from `accept_entries()` and only send the entries that are flushed

omnipaxos_core/src/storage.rs
```diff
@@ -171,6 +171,87 @@ impl<T: Entry> Snapshot<T> for () {
...
+    fn pop_front_batched_entries(&mut self, batch_size: usize) -> Vec<T> {
+        let mut entries = Vec::new();
+        for _ in 0..batch_size {
+            entries.push(self.batched_entries.remove(0));
+        }
+        entries
+    }
+    // fn clear_batched_entries(&mut self) {
+    //     self.batched_entries.clear();
+    // }
```
haraldng, line 240: We will need this function when we change leader (then the batch should be cleared)

```suggestion
     fn clear_batched_entries(&mut self) {
         self.batched_entries.clear();
     }
```

omnipaxos_core/src/sequence_paxos/follower.rs
haraldng, line 1: A follower must clear the batch whenever the leader is changed so here: https://github.com/haraldng/omnipaxos/blob/e17d4dacd922af9d3a50a21868a8590e1bd9c4b0/omnipaxos_core/src/sequence_paxos/follower.rs#L18

omnipaxos_core/src/sequence_paxos/leader.rs
haraldng, line 1: The leader must also clear its batch when it first starts its leadership here: https://github.com/haraldng/omnipaxos/blob/e17d4dacd922af9d3a50a21868a8590e1bd9c4b0/omnipaxos_core/src/sequence_paxos/leader.rs#L33

omnipaxos_core/src/sequence_paxos/mod.rs
haraldng, line 1: servers must clear batch when they reconnect with others: https://github.com/haraldng/omnipaxos/blob/e17d4dacd922af9d3a50a21868a8590e1bd9c4b0/omnipaxos_core/src/sequence_paxos/mod.rs#L348

omnipaxos_core/tests/config/test.toml
```diff
@@ -44,3 +44,13 @@ num_proposals = 10000
 gc_idx = 1000
 storage_type = "Memory"
 
+[batching_test]
+wait_timeout_ms = 3000
+num_threads= 8
+num_nodes= 3
+batch_size= 31
```
haraldng, line 51:
```suggestion
batch_size= 31   // prime number is used to check that proposals are getting decided one batch at a time.
```

omnipaxos_core/src/storage.rs
```diff
@@ -190,12 +273,16 @@ where
...
+    pub(crate) fn with(storage: I, batch_size: usize) -> Self {
+        let mut internal_store = InternalStorage {
             storage,
+            state_cache: StateCache::new(),
+            batch_size,
             _t: Default::default(),
             _i: Default::default(),
-        }
+        };
+        internal_store.load_cache();
```
haraldng, line 284: I realized that this might not work if the storage provided by the user is empty (i.e., when they create OmniPaxos for the first time). I think this could be detected if you try the test with `PersistentStorage`

To fix this, we should check if storage.get_promise() has the default ballot. If it is not, then we can load the cache.
```suggestion
        if storage.get_promised_round() != Ballot::default() {
                        internal_store.load_cache();
        }
```

haraldng (changes requested):
Good job! Most of it looks good, but there are some important things that must be fixed. I have put comments for all of them but here is a general description:
1. Move the `batch_size` to the StateCache and let all `append_` return the entries when a batch is full instead.
2. The leader should use the batching in the same way as followers, i.e., when a batch is full it will flush and then send out Accept for those flushed entries. 
3. Followers, Leaders, and all servers in general need to clear their batch (without flushing) if the leader changes or they get reconnected to some servers.
4. Check if storage has a stored a promise ballot before we try loading the cache.

I did not check the test in detail yet, but let's fix these first.

omnipaxos_core/src/sequence_paxos/leader.rs
TimXLHan:
```rust
// InternalStorage::flush_batch
pub(crate) fn flush_batch(&mut self) {
    let flushed_entries = self.state_cache.clear_batch();
    self.storage.append_entries(flushed_entries);
}
```
Is this flush function ok, or it needs to do something more?

omnipaxos_core/src/sequence_paxos/follower.rs
TimXLHan:
```rust
// InternalStorage::flush_batch
pub(crate) fn flush_batch(&mut self) {
    let flushed_entries = self.state_cache.clear_batch();
    self.storage.append_entries(flushed_entries);
}
```
To clear the batch, do we need to flush all the entries into storage or just simply discard them.

omnipaxos_core/src/sequence_paxos/mod.rs
```diff
@@ -62,28 +62,47 @@ where
...
-                    (Role::Leader, Some(v))
-                } else {
-                    (Role::Follower, None)
-                };
-                let state = (role, Phase::Accept);
-                (state, *l, lds)
+        let mut outgoing = Vec::with_capacity(BUFFER_SIZE);
+        let mut state = (Role::Follower, Phase::None);
+        let mut leader = Ballot::default();
+        let mut lds = None;
```
haraldng, line 68: Do these really need to be mutable? Can't we just initialize them as before i.e. `let (state, leader, lds) = match { ... }`

omnipaxos_core/src/storage.rs
```diff
@@ -118,7 +118,7 @@ where
     fn set_accepted_round(&mut self, na: Ballot);
 
     /// Returns the latest round in which entries have been accepted.
```
haraldng, line 120: Comment when users should return `None`

omnipaxos_core/src/storage.rs
```diff
@@ -194,8 +196,9 @@ impl<T> StateCache<T>
     where
         T: Entry
 {
-    pub fn new() -> Self {
+    pub fn new(batch_size: usize) -> Self {
         StateCache {
+            batch_size,
             batched_entries: Vec::new(),
```
haraldng, line 202: Preallocate:
```suggestion
            batched_entries: Vec::with_capacity(batch_size),
```

omnipaxos_core/src/sequence_paxos/mod.rs
```diff
@@ -62,28 +62,47 @@ where
...
+                state = (Role::Follower, Phase::Recover);
+                for peer_pid in &peers {
+                    outgoing.push(PaxosMessage {
+                        from: pid,
+                        to: *peer_pid,
+                        msg: PaxosMsg::PrepareReq,
+                    });
+                }
             }
+            // starting from scratch
```
haraldng, line 82:
```suggestion
```

omnipaxos_core/src/sequence_paxos/mod.rs
```diff
@@ -62,28 +62,47 @@ where
...
-                };
-                let state = (role, Phase::Accept);
-                (state, *l, lds)
+        let mut outgoing = Vec::with_capacity(BUFFER_SIZE);
+        let mut state = (Role::Follower, Phase::None);
+        let mut leader = Ballot::default();
+        let mut lds = None;
+        // try to do fail recovery from storage, if None then we are starting from scratch
+        match storage.get_promise() {
+            // failure recovery
```
haraldng, line 71: To be more clear, I prefer the comment inside the matching block.

omnipaxos_core/src/storage.rs
```diff
@@ -443,39 +460,58 @@ where
...
+                self.state_cache.set_decided_idx(self.storage.get_decided_idx());
+                self.state_cache.set_accepted_round(self.storage.get_accepted_round().unwrap_or_default());
+                self.state_cache.set_compacted_idx(self.storage.get_compacted_idx());
+            }
+            // set the cache to default values
+            None => {
+                self.state_cache.set_promise(Ballot::default());
+                self.state_cache.set_decided_idx(0);
+                self.state_cache.set_accepted_round(Ballot::default());
+                self.state_cache.set_compacted_idx(0);
```
haraldng, line 477: Is this needed? The cache should already be initialized with default values right?

haraldng (changes requested):

omnipaxos_core/src/storage.rs
```diff
@@ -250,6 +235,40 @@ impl<T> StateCache<T>
...
+    fn flush_if_batch_full(&mut self) -> Option<Vec<T>> {
+        let mut flushed_entries = Vec::new();
+        while self.batched_entries.len() >= self.batch_size {
+            flushed_entries.extend(self.pop_front_batched_entries());
+        }
+        if !flushed_entries.is_empty(){
+            Some(flushed_entries)
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
TimXLHan: Some times leader send `acceptdecide` or `acceptsync` with large number of log entries, under this condition we need to pop up entries from batching for serval times. For example when leader send `acceptsync(acc_of_len_31)` and `follower.batch_size == 10`.

Now I changed to calculate the capacity like this:
```rust
let mut flushed_entries = Vec::with_capacity((self.batched_entries.len() / self.batch_size) * self.batch_size);
```
Is it proper?

omnipaxos_core/src/storage.rs
```diff
@@ -233,27 +239,27 @@ impl<T> StateCache<T>
...
-    // Flushes the batched entries if the batch is full, and returns the flushed entries.
-    fn flush_if_batch_full(&mut self) -> Option<Vec<T>> {
-        let mut flushed_entries = Vec::with_capacity((self.batched_entries.len() / self.batch_size) * self.batch_size);
-        while self.batched_entries.len() >= self.batch_size {
-            flushed_entries.extend(self.batched_entries.drain(0..self.batch_size).collect::<Vec<T>>());
-        }
-        if !flushed_entries.is_empty(){
-            Some(flushed_entries)
+    // Return batched entries if the batch is full that need to be flushed in to storage.
+    fn pop_batched_entries_tobe_flushed(&mut self) -> Option<Vec<T>> {
```
haraldng, line 259:
```suggestion
    fn take_entries_to_be_flushed(&mut self) -> Option<Vec<T>> {
```

omnipaxos_core/src/storage.rs
```diff
@@ -233,27 +239,27 @@ impl<T> StateCache<T>
...
+        let flushed_entries_len = (self.batched_entries.len() / self.batch_size) * self.batch_size;
+        if flushed_entries_len > 0 {
```
haraldng, line 261: Is this calculation really needed? Isn't `flushed_entries_len` always the self.batch_size?

omnipaxos_core/src/storage.rs
```diff
@@ -468,62 +475,52 @@ where
...
-            None
-        }
+        append_res.map(|flushed_entries| self.append_entries_without_batching(flushed_entries))
     }
 
     // Append entries without batching, return the accepted index
     pub(crate) fn append_entries_without_batching(&mut self, entries: Vec<T>) -> u64 {
         self.storage.append_entries(entries);
+        self.update_accepted_idx();
         self.get_accepted_idx()
```
haraldng: The `updated_accepted_idx()` function reads from the persistent storage to get the log length. This is slow and unnecessary since we already get back the real_log length from `self.storage.append_entries(entries)`:

```suggestion
        let real_log_len = self.storage.append_entries(entries);
        let accepted_idx = real_log_len + self.state_cache.get_compacted_idx();
        self.state_cache.accepted_idx = accepted_idx;
        accepted_idx
```

omnipaxos_core/src/storage.rs
```diff
@@ -468,62 +475,52 @@ where
...
     pub(crate) fn append_on_decided_prefix(&mut self, entries: Vec<T>) -> u64 {
         let decided_idx = self.storage.get_decided_idx();
         let compacted_idx = self.get_compacted_idx();
-        self.storage
+        let res = self.storage
             .append_on_prefix(decided_idx - compacted_idx, entries)
-            + compacted_idx
+            + compacted_idx;
+        self.update_accepted_idx();
+        res
```
haraldng: Same here, see previous comment

omnipaxos_core/src/storage.rs
```diff
@@ -560,9 +557,14 @@ where
         self.storage.get_entries(from_sfx_idx, to_sfx_idx)
     }
 
+    /// Update the accepted index in the state cache
+    fn update_accepted_idx(&mut self) {
+        self.state_cache.set_accepted_idx(self.get_real_log_len() + self.get_compacted_idx());
+    }
+
```
haraldng, line 564: Remove this function. It is more clear to directly do `self.state_cache.set_accepted_idx()` and give the compacted idx manually. 

In general, please remove all these wrapper functions with a single line that just `update` or `read` from the cache. You might need to keep some of them that are being called from `SequencePaxos`, but try removing the rest. They make it confusing since they are inside InternalStorage, but are actually updating and reading from the cache.

omnipaxos_core/src/storage.rs
haraldng: Double check everywhere you do this `self.update_accepted_idx();`

haraldng (changes requested):
Please fix the way we update the accepted index which should not read from storage an extra time (see comment). Also remove any unnecessary wrapper functions in `InternalStorage`

haraldng (changes requested):
Please fix the way we update the accepted index which should not read from storage an extra time (see comment). Also remove any unnecessary wrapper functions in `InternalStorage`

omnipaxos_core/src/storage.rs
TimXLHan: If `batched_entries` == 35 and `batch_size` == 10, the `flushed_entries_len` will be 30 (not 10).

TimXLHan:
Now there is a field in `state_cache` called `real_log_len` to keep track of the length of the logs. Then all the operations that try to get real log length will go through `state_cache`.

omnipaxos_core/src/storage.rs
haraldng: Why would we want to flush 30 entries in this case? To me, it makes more sense to either flush all 35 entries or the 10 entries that the user set as batch size.

omnipaxos_core/src/storage.rs
```diff
@@ -171,6 +172,107 @@ impl<T: Entry> Snapshot<T> for () {
...
+    }
+    fn set_decided_idx(&mut self, ld: u64) {
+        self.ld = ld;
+    }
+    fn set_compacted_idx(&mut self, compacted_idx: u64) {
+        self.compacted_idx = compacted_idx;
+    }
+    fn set_real_log_len(&mut self, real_log_len: u64) {
+        self.real_log_len = real_log_len;
+    }
```
haraldng, line 246: Remove all these functions... we can set/read from them directly.

omnipaxos_core/src/storage.rs
haraldng: For simplicity, if we go over the batch size, then let's just flush all entries.

omnipaxos_core/src/storage.rs
```diff
@@ -171,6 +172,107 @@ impl<T: Entry> Snapshot<T> for () {
...
+        let flushed_entries_len = (self.batched_entries.len() / self.batch_size) * self.batch_size;
+        if flushed_entries_len > 0 {
+            Some(self.batched_entries.drain(0..flushed_entries_len).collect::<Vec<T>>())
+        } else {
+            None
+        }
+    }
+    // Clears the batched entries and returns the cleared entries. If the batch is empty,
+    // return an empty vector.
+    fn take_batch(&mut self) -> Vec<T> {
```
haraldng, line 271:
```suggestion
    fn take_batched_entries(&mut self) -> Vec<T> {
```

omnipaxos_core/src/storage.rs
```diff
@@ -644,7 +629,6 @@ where
             if idx <= decided_idx {
                 self.storage.trim(idx - compacted_idx);
                 self.set_compacted_idx(idx);
-                self.update_accepted_idx();
```
haraldng: don't we need to set real_log_len here?

haraldng (changes requested):

TimXLHan:
Now we take all the entries if batch full, then changed the `batch_test` to this: `assert!(decided_idx - last_decided_idx >= batch_size);`

omnipaxos_core/src/storage.rs
```diff
@@ -210,65 +210,38 @@ impl<T> StateCache<T>
...
-        self.take_entries_to_be_flushed()
+        self.take_entries_if_batch_full()
     }
+
     // Return batched entries if the batch is full that need to be flushed in to storage.
-    fn take_entries_to_be_flushed(&mut self) -> Option<Vec<T>> {
-        let flushed_entries_len = (self.batched_entries.len() / self.batch_size) * self.batch_size;
-        if flushed_entries_len > 0 {
-            Some(self.batched_entries.drain(0..flushed_entries_len).collect::<Vec<T>>())
+    fn take_entries_if_batch_full(&mut self) -> Option<Vec<T>> {
```
haraldng, line 234:
```suggestion
    fn take_entries_if_batch_is_full(&mut self) -> Option<Vec<T>> {
```

omnipaxos_core/src/storage.rs
```diff
@@ -171,6 +172,80 @@ impl<T: Entry> Snapshot<T> for () {
...
+    /// The maximum number of entries to batch.
+    batch_size: usize,
+    /// Vector which contains all the logged entries in-memory.
+    batched_entries: Vec<T>,
+    /// Last promised round.
+    n_prom: Ballot,
+    /// Last accepted round.
+    acc_round: Ballot,
+    /// Length of the decided log.
+    ld: u64,
```
haraldng, line 190:
```suggestion
    promise: Ballot,
    /// Last accepted round.
    accepted_round: Ballot,
    /// Length of the decided log.
    decided_idx: u64,
```

haraldng (changes requested):

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -144,15 +134,38 @@ where
...
+                        .internal_storage
+                        .append_entries_without_batching(accsync.suffix);
+                    // manually rollback set_snapshot if append suffix fails
+                    if let Err(_e) = &accepted_res {
+                        self.internal_storage
+                            .try_trim(self.internal_storage.get_accepted_idx())
+                            .expect("storage error while trying to trim log entries");
+                        self.internal_storage
+                            .append_entries_without_batching(old_log)
+                            .expect("storage error while trying to rollback log entries");
```
haraldng, line 147: If I understand correctly, we first trim because we essentially want to clear the log before appending the old entries.  To make this a bit less messy/avoid confusion I think both the trimming and then appending the old log should be hidden from here. I propose we create a new function called `rollback_log()` in InternalStorage.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -144,15 +134,38 @@ where
...
+                                .set_snapshot(old_compacted_idx, old_snapshot)
+                                .expect("storage error while trying to rollback snapshot");
+                        } else {
+                            self.internal_storage
+                                .set_compacted_idx(old_compacted_idx)
+                                .expect("storage error while trying to rollback compacted index");
+                            self.internal_storage
+                                .reset_snapshot()
+                                .expect("storage error while trying to reset snapshot");
+                        }
```
haraldng, line 159: Same here, add a new function called `rollback_snapshot(s: Option<(Snapshot, u64)>` in InternalStorage that does this.

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -455,10 +451,38 @@ where
...
+                            } else {
+                                self.internal_storage
+                                    .set_compacted_idx(old_compacted_idx)
+                                    .expect(
+                                        "storage error while trying to rollback compacted index",
+                                    );
+                                self.internal_storage
+                                    .reset_snapshot()
+                                    .expect("storage error while trying to reset snapshot");
+                            }
```
haraldng, line 478: Move this rollback logic to internal storage.

haraldng (changes requested):
Looks good. Just need to move the rollback logic of snapshot and log entries to InternalStorage to keep the code more clean.

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -456,26 +459,9 @@ where
                             .append_entries_without_batching(suffix);
                         // manually rollback set_snapshot if append suffix fails
```
haraldng, line 460:
```suggestion
                        // manually rollback snapshot and log if append suffix fails
```

omnipaxos/src/storage.rs
```diff
@@ -309,6 +309,13 @@ where
         }
     }
 
+    pub(crate) fn rollback_and_panic(&mut self, values: Vec<RollbackValue>, msg: &str) {
```
haraldng, line 312: Let's use this function when possible, for example [here](https://github.com/haraldng/omnipaxos/blob/c883cc9c25d03f047820c016d761f6570b8b4404/omnipaxos/src/sequence_paxos/follower.rs#L259) and [here](https://github.com/haraldng/omnipaxos/blob/c883cc9c25d03f047820c016d761f6570b8b4404/omnipaxos/src/sequence_paxos/follower.rs#L259). (Please check other places if we can replace it)

omnipaxos/src/storage.rs
```diff
@@ -326,6 +333,27 @@ where
         }
     }
 
+    /// Rollback the log in the storage using given log entries.
+    pub(crate) fn rollback_log(&mut self, entries: Vec<T>) {
+        self.try_trim(self.get_accepted_idx())
+            .expect("storage error while trying to trim log entries");
```
haraldng, line 339:
```suggestion
            .expect("storage error while trying to trim log entries before rolling back");
```

omnipaxos/src/omni_paxos.rs
```diff
@@ -27,7 +27,7 @@ use toml;
 /// * `configuration_id`: The identifier for the configuration that this Sequence Paxos replica is part of.
 /// * `pid`: The unique identifier of this node. Must not be 0.
 /// * `peers`: The peers of this node i.e. the `pid`s of the other replicas in the configuration.
-/// * `batch_size`: The size of the buffer for log batching. 1 means no batching.
+/// * `batch_size`: The size of the buffer for log batching. default is 1, means no batching.
```
haraldng, line 30:
```suggestion
/// * `batch_size`: The size of the buffer for log batching. The default is 1, which means no batching.
```

haraldng (changes requested):
Use rollback_and_panic whenever it's possible and update docs/comments.

haraldng (approved):

Issue #68, opened 2023-04-19
Complete reconnection automation


kevin-harrison:
Together issue #61  and issue #67 enable us to discover and react to dropped `Prepare`, `AcceptSync`, `AcceptDecide`, and `Decide` messages rather than relying on the user to call reconnected().

They don't cover cases of dropped `AcceptStopSign`, `DecideStopSign`, `PrepareReq` messages. The absence of these messages can't be easily detected with sequence numbers. Instead they can be resent after some timeout (can for example use the same timeout as BLE).

PR #85, opened 2023-04-30
Feature tick API

kevin-harrison:

## Issues
Addresses Issue #68. Makes OmniPaxos resistant to network errors resulting in dropped, duplicated, and out-of-order messages.

This is achieved by:
1. Periodically resending messages in the prepare phase (`Prepare` and `Promise`) if they don't receive the expected response from the receiver.
2. Giving sequence numbers to messages in the accept phase and detecting sequence breaks (see issue #61).
3. Dropped messages in the accept phase which can't always be detected are `AcceptStopSign` (if network errors prevent it from becoming decided) and `DecideStopSign` (as its the end of the sequence). `AcceptStopSign` is solved by resending, but there is no current solution for dropped `DecideStopSign` messages.

## Breaking Changes
Driving leader election is done with the `tick()` function instead of `election_timeout()`.


omnipaxos_core/src/sequence_paxos/leader.rs
```diff
@@ -48,19 +48,8 @@ where
...
-            };
-            /* send prepare */
-            for pid in &self.peers {
-                self.outgoing.push(PaxosMessage {
-                    from: self.pid,
-                    to: *pid,
-                    msg: PaxosMsg::Prepare(prep),
-                });
+            for pid in self.peers.clone() {
+                self.send_prepare(pid);
```
haraldng, line 52: Why cloning? I think it should work by dereferencing `*pid` like we did before.

omnipaxos_core/src/sequence_paxos/mod.rs
```diff
@@ -213,6 +214,43 @@ where
...
+                            self.send_accept_stopsign(follower, ss.stopsign.clone());
+                        }
+                    }
+                }
+                _ => (),
+            }
+        }
+
+        // Resend PrepareReq
+        if self.state == (Role::Follower, Phase::Recover) {
```
haraldng, line 245: Let's pattern match directly on `self.state` something like this:

```
match self.state {
       (Role::Leader, phase) => {
                  // resend prepare
                  // resend stopsign if there is a pending stopsign and phase == Phase::accept. 
       }, 
      (Role::Follower, Phase::Recover) => ...
}
```

omnipaxos_core/src/util.rs
```diff
@@ -147,6 +147,10 @@ where
         self.accepted_stopsign[Self::pid_to_idx(from)] = true;
     }
 
+    pub fn get_accepted_stopsign(&mut self, from: u64) -> bool {
```
haraldng, line 150: For functions that return `bool`, a good convention is to name them as questions that can be answered with true/false :D

```suggestion
    pub fn follower_has_accepted_stopsign(&mut self, pid: NodeId) -> bool {
```

omnipaxos_core/src/omni_paxos.rs
```diff
@@ -81,6 +89,8 @@ impl Default for OmniPaxosConfig {
             pid: 0,
             peers: Vec::new(),
             buffer_size: BUFFER_SIZE,
+            election_tick_timeout: 100,
+            resend_message_tick_timeout: 500,
```
haraldng, line 93: Define these as constants. Not sure where BUFFER_SIZE is right now but all default values should be placed together somewhere where it makes sense

haraldng (changes requested):

omnipaxos_core/src/sequence_paxos/leader.rs
kevin-harrison: `send_prepare` requires a mutable reference to self since it mutates `self.outgoing`. The compiler is worried that we immutably borrow `self.peers` and then do a mutable borrow since it can't tell that we wont mutate `self.peers` in the `send_prepare` function call. https://smallcultfollowing.com/babysteps/blog/2018/11/01/after-nll-interprocedural-conflicts/
```
error[E0502]: cannot borrow `*self` as mutable because it is also borrowed as immutable
  --> omnipaxos_core/src/sequence_paxos/leader.rs:52:17
   |
51 |             for pid in &self.peers {
   |                        -----------
   |                        |
   |                        immutable borrow occurs here
   |                        immutable borrow later used here
52 |                 self.send_prepare(*pid);
   |                 ^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here

```

omnipaxos_core/src/omni_paxos.rs
```diff
@@ -224,6 +237,18 @@ where
...
 
+    /// Drives the election process (see `election_timeout()`) every `election_tick_timeout`
+    /// ticks. Also drives the detection and resending of dropped omnipaxos messages every `resend_message_tick_timeout` ticks.
+    pub fn tick(&mut self) {
+        self.tick += 1;
+        if self.tick % self.election_tick_timeout == 0 {
+            self.election_timeout();
+        }
+        if self.tick % self.resend_message_tick_timeout == 0 {
+            self.seq_paxos.resend_message_timeout();
```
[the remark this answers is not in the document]
kevin-harrison: I initially reset the tick to 0 when the tick equaled the product of the timeouts, but then I realized if you tick once every millisecond you will have enough ticks for 584942417 years. But a `Logical Clock` struct could be nicer. We have a similar problem with the sequence numbers which could also overflow.

omnipaxos_core/src/sequence_paxos/leader.rs
haraldng: Ok since this is in the prepare phase i.e., it is performed rarely it should be fine.

omnipaxos_core/src/omni_paxos.rs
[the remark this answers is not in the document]
haraldng: There is no reason to leave the possibility of overflow if there is a way to prevent it. With sequence numbers, I don't think there is a better way, and it also gets reset per leader change. It would be amazing if OmniPaxos didn't even change leader for that many years :D

omnipaxos_core/src/util.rs
```diff
@@ -174,6 +178,15 @@ where
...
     }
 
+    pub fn get_unpromised_peers(&self) -> Vec<u64> {
+        self.decided_indexes
+            .iter()
+            .enumerate()
+            .filter(|(pid, x)| x.is_none() && *pid != Self::pid_to_idx(self.n_leader.pid))
+            .map(|(idx, _)| (idx + 1) as u64)
+            .collect()
+    }
```
haraldng, line 188:
```suggestion
    pub fn get_unpromised_peers(&self) -> Vec<NodeId> {
        self.decided_indexes
            .iter()
            .enumerate()
            .filter(|(pid, x)| x.is_none() && *pid != Self::pid_to_idx(self.n_leader.pid))
            .map(|(idx, _)| (idx + 1) as NodeId)
            .collect()
    }
```

haraldng: I had made the same mistake for `get_promised_followers()`. Could please also update it there?

omnipaxos_core/src/omni_paxos.rs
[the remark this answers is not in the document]
kevin-harrison: We could use the overflow as our mechanism for resetting the ticks. I think unsigned int overflow is defined behavior in Rust? Then every 1/2 billion-ish years your election timeout and resending messages may get triggered earlier than expected, but nothing breaks.

This can solve the sequence number overflow problem aswell. TCP sequence numbers do this [wrap-around](https://www.geeksforgeeks.org/wrap-around-concept-and-tcp-sequence-number/) actually.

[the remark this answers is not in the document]
haraldng: No, just reset the logical clock to 0 when it reaches its timeout. If you want we can fix the wrap-around for message seq. numbers in a future PR.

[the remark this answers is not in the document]
haraldng: Using custom handling for overflow is an overkill for this case and just makes the code less intuitive.

omnipaxos_core/src/sequence_paxos/mod.rs
haraldng: I'm not sure I agree. It is more readable because you first only match on the `Role` (which is in fact either `Leader` or `Follower`) while in your original code, we have three ifs even though some of them should be if/else.

I also don't see a problem with matching `() => _`. You can also get rid of one of them by doing `if let Some(...)` when checking the `StopSign`.

Two side notes:
1. Do we need to `clone` the promise? Ballots should be `copy`, and if we really need to clone it, we should ideally do it inside the `match` or `if`.
2. I think the servers are initialized to be in the Prepare phase, so it could be that we're in the Prepare phase without having a promise which can cause unnecessary logging.

omnipaxos_core/src/sequence_paxos/follower.rs
```diff
@@ -181,12 +180,10 @@ where
             let msg_status = self.current_seq_num.check_msg_status(acc_ss.seq_num);
             match msg_status {
                 MessageStatus::First => {
-                    #[cfg(feature = "logging")]
-                    warn!(
-                        self.logger,
-                        "AcceptStopSign cannot be the first message in a sequence!"
-                    );
-                    return;
+                    // psuedo-AcceptSync for reconfigurations
```
haraldng, line 183:
```suggestion
                    // psuedo-AcceptSync for when using predetermined leader without prepare phase.
```

omnipaxos_core/src/sequence_paxos/mod.rs
```diff
@@ -217,37 +218,60 @@ where
...
+                        // Resend Promise
+                        match &self.cached_promise {
+                            Some(promise) => {
+                                self.outgoing.push(PaxosMessage {
+                                    from: self.pid,
+                                    to: promise.n.pid,
+                                    msg: PaxosMsg::Promise(promise.clone()),
+                                });
+                            },
+                            None => {
```
haraldng, line 263: This shouldn't happen, but if it really does, perhaps it makes sense to try and recover this server by going into recovery mode and sending PrepareReq to all peers. This will also help to avoid excessive logging.

omnipaxos_core/src/util.rs
```diff
@@ -336,3 +342,29 @@ impl SequenceNumber {
...
+    pub fn tick(&mut self) {
+        self.time += 1;
+        if self.time == self.timeout {
+            self.time = 0;
+        }
+    }
+
+    pub fn reached_timeout(&self) -> bool {
+        self.time == 0
+    }
```
haraldng, line 369: Since we always check the timeout after each tick, I think a better API is to have one function `tick_and_check_timeout() -> bool`. This way the rust compiler will also complain if we call it and disregard the result (which will help us avoiding mistakes when coding).

omnipaxos/src/util.rs
```diff
@@ -338,28 +338,22 @@ impl SequenceNumber {
...
         self.time += 1;
         if self.time == self.timeout {
             self.time = 0;
+            return true;
         }
-    }
-
-    pub fn reached_timeout(&self) -> bool {
-        self.time == 0
+        false
```
haraldng: Can remove return and do: 
```
if ... { 
     ...
     true
} else {
     false
}
```

haraldng (commented):
Looks good! Only a few minor comments.

Do we need new tests for this?

omnipaxos/tests/config/node1.toml
```diff
@@ -2,8 +2,10 @@
 configuration_id = 1
 pid = 1
 peers = [2, 3]
-buffer_size = 100000
+buffer_size = 100001
+election_tick_timeout = 32
+resend_message_tick_timeout = 6343
```
haraldng, line 7: Comment why we use these values

haraldng (approved):

Issue #62, opened 2023-04-18
Flexible quorums


haraldng:
Allow users to configure the quorums so that instead of majority quorums they could use any prepare and accept phase quorums sizes: `p` and `a` as long as at least one node from any accept phase quorum will be in the prepare phase quorum.

Example:
`N = 5, p=4, a=2`.

An entry is decided when 2 nodes have accepted but any prepare phase requires instead 4 promises before proceeding to accept phase. 

More resources: https://fpaxos.github.io/

haraldng:
Closed with #91

PR #91, opened 2023-05-14
Flexible Quorums

kevin-harrison:

## Issues
Addresses issue #62.  Adds flexible quorums options to `OmniPaxosConfig`. A node can become a leader if it can form a prepare quorum, and can retain leadership so long as it can form an accept quorum.

In cases of partial connectivity a leader can lose leadership despite forming an accept quorum. This can happen if another node is prepare-quorum-connected after losing connection to the accept-quorum-connected leader.

## Breaking Changes
 - `OmniPaxos::fail_recovery()` is no long callable as it is no longer needed
 - `OmniPaxosConfg` no longer has an `initial_leader` field

## Other Changes
 - `TestSystem::with()` now takes a `TestConfig` as an argument
 - Fixes `TestSystem::get_elected_leader()` deadlock

### DIFF
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -1,5 +1,5 @@
 
-The OmniPaxos library is mainly driven by the [**OmniPaxos**](omnipaxos/index.md) struct.  It is a plain Rust struct and the user therefore needs to provide a network implementation themselves to actually send and receive messages. In this tutorial we will show how a user should interact with this struct in order to implement a strongly consistent, replicated log. This tutorial will focus on how to use the library and showcase its features. 
+The OmniPaxos library is mainly driven by the [**OmniPaxos**](omnipaxos/index.md) struct. It is a plain Rust struct and the user therefore needs to provide a network implementation themselves to actually send and receive messages. In this tutorial we will show how a user should interact with this struct in order to implement a strongly consistent, replicated log. This tutorial will focus on how to use the library and showcase its features.
 <!-- For the properties and advantages of OmniPaxos in comparison to other similar protocols, we refer to the Omni-Paxos paper. -->
 
 ## Bleeding Edge
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -1,5 +1,5 @@
 
-The OmniPaxos library is mainly driven by the [**OmniPaxos**](omnipaxos/index.md) struct. It is a plain Rust struct and the user therefore needs to provide a network implementation themselves to actually send and receive messages. In this tutorial we will show how a user should interact with this struct in order to implement a strongly consistent, replicated log. This tutorial will focus on how to use the library and showcase its features.
+The OmniPaxos library is mainly driven by the [**OmniPaxos**](omnipaxos/index) struct. It is a plain Rust struct and the user therefore needs to provide a network implementation themselves to actually send and receive messages. In this tutorial we will show how a user should interact with this struct in order to implement a strongly consistent, replicated log. This tutorial will focus on how to use the library and showcase its features.
 <!-- For the properties and advantages of OmniPaxos in comparison to other similar protocols, we refer to the Omni-Paxos paper. -->
 
 ## Bleeding Edge
--- a/docs/omnipaxos/leader_election.md
+++ b/docs/omnipaxos/leader_election.md
@@ -10,3 +10,29 @@ omni_paxos.election_timeout();
 If a leader has failed, it will be detected in one election timeout and a new leader will be elected in the next timeout (if possible).
 
 > **Note:** The `leader_priority` field in `OmniPaxosConfig` allows user to give desired servers a higher priority to get elected upon a leader change.
+
+## Flexible Quorums
+OmniPaxos also offers support for flexible quorums. Normally a node needs to consult a majority of nodes to become a leader and then as a leader consult a majority of nodes to append to the log. Flexible quorums allow for a trade off between the number of nodes that need to be consulted in each of these two scenarios.
+
+OmniPaxos can be configured to use a flexible quorum as follows:
+```rust,edition2018,no_run,noplaypen
+use omnipaxos::{OmniPaxosConfig, util::FlexibleQuroum};
+
+let flex_quorum = FlexibleQuorum {
+    read_quorum_size: 5,
+    write_quorum_size: 3,
+};
+
+let config = OmniPaxosConfig {
+    configuration_id: 1,
+    pid: 1,
+    peers: [2, 3, 4, 5, 6, 7],
+    flexible_quorum: Some(flex_quorum),
+    ..Default::deafult()
+};
+```
+In OmniPaxos, becoming a leader involves reading from the log and so `read_quorum_size` is the number of nodes to consult in order to become a leader. The `write_quorum_size` is the number of nodes to consult in order to append to the log.
+
+Appending to the log is a much more common occurance than electing a leader so reducing the `write_quorum_size` can reduce the latency and sometimes even increase throughput. Note, however, that reducing the `write_quorum_size` also decreases the fault tolerance of the cluster to `write_quorum_size - 1`.
+
+In order to guarantee safety the `read_quorum_size` and `write_quorum_size` must overlap (i.e. `read_quorum_size` + `write_quorum_size` > # of nodes in cluster). In practice this means that reducing the `write_quorum_size` requires increasing the `read_quorum_size`.
--- a/docs/omnipaxos/leader_election.md
+++ b/docs/omnipaxos/leader_election.md
@@ -26,7 +26,7 @@ let flex_quorum = FlexibleQuorum {
 let config = OmniPaxosConfig {
     configuration_id: 1,
     pid: 1,
-    peers: [2, 3, 4, 5, 6, 7],
+    peers: vec![2, 3, 4, 5, 6, 7],
     flexible_quorum: Some(flex_quorum),
     ..Default::deafult()
 };
--- a/docs/omnipaxos/leader_election.md
+++ b/docs/omnipaxos/leader_election.md
@@ -1,21 +1,25 @@
 A unique feature of Omni-Paxos is the guaranteed progress with one quorum-connected server using Ballot Leader Election (BLE). In this section, we will see how the leader election is used in OmniPaxos.
 
-To detect any leader failures and elect a new leader, you must call the function ``election_timeout()`` periodically (e.g., every 100ms).
+To detect any leader failures and elect a new leader, you must call the function ``tick()`` periodically. The number of ticks required before an election timeout is triggered can be configured with the ``election_tick_timeout`` field of ``OmniPaxosConfig``. For example, if ``tick()`` is called every 10ms and the ``election_tick_timeout`` is 10, then an election will be triggered every 100ms.
 
-```
+```rust
 // Call this periodically
-omni_paxos.election_timeout();
+omni_paxos.tick();
 ```
 
 If a leader has failed, it will be detected in one election timeout and a new leader will be elected in the next timeout (if possible).
 
 > **Note:** The `leader_priority` field in `OmniPaxosConfig` allows user to give desired servers a higher priority to get elected upon a leader change.
 
+In some cases, a network error will require the resending of messsages between leader and follower. The `tick()` function also drives this behavior. The number of ticks required before checking if a message needs to be resent can be configured with the ``resend_message_tick_timeout`` field of ``OmniPaxosConfig``.
+
+
+
 ## Flexible Quorums
 OmniPaxos also offers support for flexible quorums. Normally a node needs to consult a majority of nodes to become a leader and then as a leader consult a majority of nodes to append to the log. Flexible quorums allow for a trade off between the number of nodes that need to be consulted in each of these two scenarios.
 
 OmniPaxos can be configured to use a flexible quorum as follows:
-```rust,edition2018,no_run,noplaypen
+```rust
 use omnipaxos::{OmniPaxosConfig, util::FlexibleQuroum};
 
 let flex_quorum = FlexibleQuorum {
@@ -36,3 +40,4 @@ In OmniPaxos, becoming a leader involves reading from the log and so `read_quoru
 Appending to the log is a much more common occurance than electing a leader so reducing the `write_quorum_size` can reduce the latency and sometimes even increase throughput. Note, however, that reducing the `write_quorum_size` also decreases the fault tolerance of the cluster to `write_quorum_size - 1`.
 
 In order to guarantee safety the `read_quorum_size` and `write_quorum_size` must overlap (i.e. `read_quorum_size` + `write_quorum_size` > # of nodes in cluster). In practice this means that reducing the `write_quorum_size` requires increasing the `read_quorum_size`.
+
--- a/docs/omnipaxos/leader_election.md
+++ b/docs/omnipaxos/leader_election.md
@@ -12,32 +12,3 @@ If a leader has failed, it will be detected in one election timeout and a new le
 > **Note:** The `leader_priority` field in `OmniPaxosConfig` allows user to give desired servers a higher priority to get elected upon a leader change.
 
 In some cases, a network error will require the resending of messsages between leader and follower. The `tick()` function also drives this behavior. The number of ticks required before checking if a message needs to be resent can be configured with the ``resend_message_tick_timeout`` field of ``OmniPaxosConfig``.
-
-
-
-## Flexible Quorums
-OmniPaxos also offers support for flexible quorums. Normally a node needs to consult a majority of nodes to become a leader and then as a leader consult a majority of nodes to append to the log. Flexible quorums allow for a trade off between the number of nodes that need to be consulted in each of these two scenarios.
-
-OmniPaxos can be configured to use a flexible quorum as follows:
-```rust
-use omnipaxos::{OmniPaxosConfig, util::FlexibleQuroum};
-
-let flex_quorum = FlexibleQuorum {
-    read_quorum_size: 5,
-    write_quorum_size: 3,
-};
-
-let config = OmniPaxosConfig {
-    configuration_id: 1,
-    pid: 1,
-    peers: vec![2, 3, 4, 5, 6, 7],
-    flexible_quorum: Some(flex_quorum),
-    ..Default::deafult()
-};
-```
-In OmniPaxos, becoming a leader involves reading from the log and so `read_quorum_size` is the number of nodes to consult in order to become a leader. The `write_quorum_size` is the number of nodes to consult in order to append to the log.
-
-Appending to the log is a much more common occurance than electing a leader so reducing the `write_quorum_size` can reduce the latency and sometimes even increase throughput. Note, however, that reducing the `write_quorum_size` also decreases the fault tolerance of the cluster to `write_quorum_size - 1`.
-
-In order to guarantee safety the `read_quorum_size` and `write_quorum_size` must overlap (i.e. `read_quorum_size` + `write_quorum_size` > # of nodes in cluster). In practice this means that reducing the `write_quorum_size` requires increasing the `read_quorum_size`.
-
--- a/omnipaxos/src/ballot_leader_election.rs
+++ b/omnipaxos/src/ballot_leader_election.rs
@@ -18,7 +18,7 @@ use serde::{Deserialize, Serialize};
 use slog::{debug, info, trace, warn, Logger};
 
 /// Used to define a Sequence Paxos epoch
-#[derive(Clone, Copy, Eq, Debug, Default, Ord, PartialOrd, PartialEq)]
+#[derive(Clone, Copy, Eq, Debug, Default, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Ballot {
     /// Ballot number
@@ -39,14 +39,25 @@ impl Ballot {
     }
 }
 
+impl Ord for Ballot {
+    fn cmp(&self, other: &Self) -> Ordering {
+        (self.n, self.priority, self.pid).cmp(&(other.n, other.priority, other.pid))
+    }
+}
+
+impl PartialOrd for Ballot {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
 /// Used to define a BLE epoch
 #[derive(Clone, Copy, Eq, Debug, Default, PartialEq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct BLEBallot {
+struct BLEBallot {
     /// Ballot of a replica.
-    pub ballot: Ballot,
+    ballot: Ballot,
     /// Used to determine if the replica is a candidate to become a leader or remain a leader.
-    pub connectivity: u8,
+    connectivity: u8,
 }
 
 impl BLEBallot {
@@ -98,12 +109,15 @@ pub(crate) struct BallotLeaderElection {
     /// The current round of the heartbeat cycle.
     hb_round: u32,
     /// Vector which holds all the received heartbeats
-    ballots: Vec<BLEBallot>,
+    ballots: Vec<(Ballot, u8)>,
     /// Holds the current ballot of this instance.
-    current_ballot: BLEBallot,
+    current_ballot: Ballot,
+    /// The number of replicas inside the cluster that this instance is
+    /// connected to (based on heartbeats received) including itself.
+    connectivity: u8,
     /// Current elected leader.
     leader: Option<Ballot>,
-    /// The number of replicas inside the cluster whose heartbeats are needed to become and remain leader.
+    /// The number of replicas inside the cluster whose heartbeats are needed to become and remain the leader.
     quorum: Quorum,
     /// Vector which holds all the outgoing messages of the BLE instance.
     outgoing: Vec<BLEMessage>,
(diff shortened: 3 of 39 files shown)


kevin-harrison:
The flexible quorum sizes are configured with the `leader_quorum_size` and `append_quorum_size` fields. The minimal `leader_quorum_size` = `# nodes` - `append_quorum_size` + 1. Is there any advantage to allowing for a larger `leader_quorum_size`? Also allowing for larger `leader_quorum_size` can break the `append_quorum_size` - 1 fault tolerance guarantee.

kevin-harrison:
Should we allow for `append_quorum_size` = 1? Right now, with `append_quorum_size` = 1  nothing can be decided since deciding happens in `handle_accepted`, but no accepts are sent since the append quorum is simply the leader.

omnipaxos/tests/ble_test.rs
```diff
@@ -15,12 +15,7 @@ use utils::{TestConfig, TestSystem};
 fn ble_test() {
     let cfg = TestConfig::load("ble_test").expect("Test config loaded");
 
-    let mut sys = TestSystem::with(
-        cfg.num_nodes,
-        cfg.election_timeout_ms,
-        cfg.num_threads,
-        cfg.storage_type,
-    );
+    let mut sys = TestSystem::with(cfg.clone());
```
haraldng, line 18: I think clone can be avoided if you assign the num_elections and the ms timeout as variables before.

omnipaxos/src/omni_paxos.rs
```diff
@@ -62,6 +67,31 @@ impl OmniPaxosConfig {
...
+                // TODO: currently accept_quorum_size = 1 doesn't work, add to check?
+                assert!(
+                    accept_quorum_size > 0 && accept_quorum_size <= num_nodes,
+                    "Accept quorum must be in range 0 to # of nodes in the cluster"
+                );
+                assert!(prepare_quorum_size >= accept_quorum_size, "Leader election cannot guarantee progress if prepare quorum is smaller than accept quorum");
+            }
+            (None, Some(_)) => panic!("Prepare quorum must also be defined"),
+            (Some(_), None) => panic!("Accept quorum must also be defined"),
+            (None, None) => (),
```
haraldng, line 92: To make the code clean both here and in Sequence Paxos when checking quorums we should have an enum:

Quorum {
      Majority,
      Flexible(usize, usize)
}

where the default is Majority. Then we can also have convenient utility functions:

impl QuorumConfig {
     fn got_accept_quorum -> bool {}
     fn got_prepare_quorum -> bool {}
}

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -56,8 +56,8 @@ pub(crate) struct BallotLeaderElection {
     quorum_connected: bool,
     /// Current elected leader.
     leader: Option<Ballot>,
-    /// The majority of replicas inside a cluster. It is measured in ticks.
-    majority: usize,
+    /// The number of replicas inside the cluster whose vote is needed to become leader.
+    quorum_size: usize,
```
haraldng, line 60: We need to keep track of those two different quorum sizes in BLE too.

haraldng (changes requested):

omnipaxos/src/omni_paxos.rs
kevin-harrison: Maybe we have the struct without the functions and once it gets passed to `SequencePaxos::build` we go back to the `prepare_quorum_size` and `accept_quorum_size` fields.

omnipaxos/tests/ble_test.rs
kevin-harrison: Maybe `TestConfig` can just implement `Copy` since its so small, only generated for tests, and keeps it so that adding new fields doesn't require changing the start of every single test.

omnipaxos/src/omni_paxos.rs
haraldng: Good point. I would say let's keep the flexible quorum field as an Option<(usize, usize)> in OmniPaxosConfig. 

In Sequence Paxos, I think we can use `Majority(usize)` to avoid recalculations. I also still prefer the shared functions to keep the code clean.

omnipaxos/src/omni_paxos.rs
```diff
@@ -22,11 +22,9 @@ use toml;
 /// * `peers`: The peers of this node i.e. the `pid`s of the other replicas in the configuration.
 /// * `buffer_size`: The buffer size for outgoing messages.
 /// * `skip_prepare_use_leader`: The initial leader of the cluster. Could be used in combination with reconfiguration to skip the prepare phase in the new configuration.
-/// * `leader_quorum_size`: The number of votes a node needs to become the leader.
-/// * `append_quorum_size`: The number of nodes that need to participate to progress the log replication.
+/// * `flexible_quorum` : The (read_quorum_size, write_quorum_size). read_quorum_size is the number of nodes a leader needs to consult to get a synced view of the log. write_quorum_size is the number of acknowledgements a leader needs to write to the log.
```
haraldng, line 25:
```suggestion
/// * `flexible_quorum` : The (read_quorum_size, write_quorum_size). read_quorum_size is the number of nodes a leader needs to consult to get an up-to-date view of the log. write_quorum_size is the number of acknowledgments a leader needs to commit an entry to the log.
```

omnipaxos/src/omni_paxos.rs
```diff
@@ -67,31 +64,23 @@ impl OmniPaxosConfig {
...
+            );
+            assert!(
+                prepare_quorum_size >= 1 && prepare_quorum_size <= num_nodes,
+                "Read quorum must be in range 1 to # of nodes in the cluster"
+            );
+            assert!(
+                accept_quorum_size >= 2 && accept_quorum_size <= num_nodes,
+                "Write quorum must be in range 2 to # of nodes in the cluster"
+            );
+            assert!(prepare_quorum_size >= accept_quorum_size, "Leader election cannot guarantee progress if read quorum is smaller than write quorum");
```
haraldng, line 82: This last assert is not needed. We can have a smaller prepare quorum e.g., N=5, prepare_quorum=2, accept_quorum=4 is fine.

haraldng, line 82: I think we should move these asserts to the `Quorum::with()` function. This way the checks will be performed regardless of where we build them i.e. OmniPaxosConfig, SequencePaxosConfig, BLEConfig etc.

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -124,6 +108,14 @@ where
         #[cfg(feature = "logging")]
         {
             info!(paxos.logger, "Paxos component pid: {} created!", pid);
+            if let Quorum::Flexible(prepare_quorum_size, accept_quorum_size) = quorum {
+                if prepare_quorum_size > num_nodes - accept_quorum_size + 1 {
+                    warn!(
+                        paxos.logger,
+                        "Unnecessary overlaps in read and write quorums. The optimal read quorum size = (# of nodes) - (write quorum size) + 1."
```
haraldng, line 115:
```suggestion
                        "Unnecessary overlaps in read and write quorums. Read and Write quorums only need to be overlapping by one node i.e., read_quorum_size + write_quorum_size = num_nodes + 1"
```

omnipaxos/tests/config/node1.toml
```diff
@@ -6,6 +6,5 @@ buffer_size = 100000
 logger_file_path = "/omnipaxos/logs"
 leader_priority = 2
 skip_prepare_use_leader = { n = 5, priority = 2, pid = 2 }
-leader_quorum_size = 3
-append_quorum_size = 1
+flexible_quorum = [3, 1]
```
haraldng, line 9: comment which one is read and which one is write quorum.

omnipaxos/src/omni_paxos.rs
```diff
@@ -36,8 +34,7 @@ pub struct OmniPaxosConfig {
     pub peers: Vec<u64>,
     pub buffer_size: usize,
     pub skip_prepare_use_leader: Option<Ballot>,
-    pub leader_quorum_size: Option<usize>,
-    pub append_quorum_size: Option<usize>,
+    pub flexible_quorum: Option<(usize, usize)>,
```
haraldng, line 37: On a second thought, it is probably more clear to have a custom struct `FlexibleQuorum` that the user faces:
```
pub struct FlexibleQuorum {
        read_quorum_size: usize,
        write_quorum_size: usize,
}
```

```suggestion
    pub flexible_quorum: Option<FlexibleQuorum>,
```

haraldng: our Quorum enum can then use it as well `Quorum::Flexible(FlexibleQuorum)`

haraldng (changes requested):

omnipaxos/src/omni_paxos.rs
kevin-harrison: If we allow this then the flexible quorum fault tolerance of `accept_quorum - 1` no longer holds. In your example,  the flexible quorum should have a fault tolerance of 4-1=3 but even just 2 nodes failing means that an accept quorum cannot be formed.

But the error message doesn't make sense since the updated BLE that takes into account both quorum sizes. Maybe the check should just be what they have in the flexible quorum paper `accept_quorum_size < num_nodes / 2`? Otherwise the fault tolerance guarantee becomes `min(accept_q - 1, num_nodes - accept_q)` I think.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -138,19 +140,16 @@ impl BallotLeaderElection {
...
-                    }
-                },
-            )
+            .filter_map(|(ballot, connectivity)| {
+                if self.quorum.is_accept_quorum(connectivity) {
+                    Some(ballot)
+                } else {
+                    None
+                }
+            })
```
haraldng, line 152: Since we now have the actual connectivity, it would be nice to also prioritize the nodes that have better connectivity. Right now, the `max()` function will compare the ballots lexicographically, but since connectivity is not part of the `Ballot` we probably need to map a Ballot `b`  into a tuple `(b.n, b.priority, connectivity, b.pid)` before picking out the max.

haraldng (changes requested):
Looks good, the comments are mostly related to new ideas and improvements I found while reviewing. Let me know if anything is unclear.

omnipaxos/src/messages.rs
```diff
@@ -224,8 +224,8 @@ pub mod ballot_leader_election {
         pub round: u32,
         /// Ballot of a replica.
         pub ballot: Ballot,
-        /// States if the replica is a candidate to become a leader.
-        pub quorum_connected: bool,
+        /// Used to determine if the replica is a candidate to become a leader or remain a leader.
+        pub connectivity: usize,
```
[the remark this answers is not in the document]
kevin-harrison: This makes the maximum number of nodes a cluster can have be 255.  As far as performance `u8`s still have to go into 32/64bit registers in the cpu. There is maybe an argument for better SIMD/compiler optimizations or CPU cache space, but I don't think this small of an optimization is worth the limitation (even if there are very few systems doing replication on that scale today).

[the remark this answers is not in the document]
kevin-harrison: Also there are currently a lot of structs in the code that use `usize` which really should only be used for indexing. Maybe we can change those to unsigned ints to be more clear about their purpose.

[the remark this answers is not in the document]
haraldng: I was referring to the overhead/size of the heartbeats as network I/O and not the CPU. But in general, there is no need to overthink these kind of things. We don't want to `usize` and want to use the smallest suitable type, therefore we should use `u8`. 

What happens in network/compiler/hardware layer is out of our control and we don't need to consider it. We do our part by using the most suitable type (`u8`). If the other layers can do optimizatons, that's fine and good. If not, we cannot do anything about it anyway.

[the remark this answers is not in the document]
haraldng: You're probably right. If you're passionate about where to use usize and unsigned ints, feel free to create an issue that describes all places you think need modifying that you or somebody else can try fixing. But from what I remember, most user-facing structs and API use unsigned ints.

omnipaxos/src/messages.rs
```diff
@@ -224,8 +225,10 @@ pub mod ballot_leader_election {
         pub round: u32,
         /// Ballot of a replica.
         pub ballot: Ballot,
-        /// States if the replica is a candidate to become a leader.
-        pub quorum_connected: bool,
+        /// Custom priority parameter
+        pub priority: u32,
+        /// Used to determine if the replica is a candidate to become a leader or remain a leader.
+        pub connectivity: u8,
```
haraldng, line 231: Use `BLEBallot` instead.

haraldng (changes requested):
Thanks! Please store `leader` as `Option<Ballot>` instead of `Option<BLEBallot>` to avoid unnecessary leader changes

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -138,25 +201,16 @@ impl BallotLeaderElection {
...
-                    } else {
-                        None
-                    }
-                },
-            )
+            .filter(|ballot| self.quorum.is_accept_quorum(ballot.connectivity as usize))
             .max()
             .unwrap_or_default();
 
         if top_ballot < self.leader.unwrap_or_default() {
```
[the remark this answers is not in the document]
kevin-harrison: Do we not run into problems with the priority then? For example
```rust
// New incoming top_ballot
let top_ballot = BLEBallot {
    n: 1,
    priority: 0,
    connectivity: 5,
    pid: 2,
};
// The full BLEBallot of the leader when it was elected
let leader = BLEBallot {
    n: 1,
    priority: 5,
    connectivity: 5,
    pid: 1,
}
```
The `top_ballot` shouldn't become leader but it will be considered greater because we ignore the priority fields (Due to to storing `self.leader` only as a `Ballot`)?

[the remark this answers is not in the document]
haraldng: After thinking more about this, I think it will be hard to try and ignore the priority field. Because unlike with connectivity where we do a filter, the priority is used to order ballots. And if we would ignore the priority sometimes, it would mean ignoring a higher ballot, which is incorrect (the QLE properties state that we eventually should agree on the highest ballot).

I would like to hear what you think, but I believe the best is to revert back to having `priority` as a field in `Ballot` and have `leader: Ballot` in BLE. This will mean that if a priority changes, that could also change leadership. So we have to make sure that we make this clear in the documentation/tutorial so that the users know that.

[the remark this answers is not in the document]
haraldng: Another thing that is less trivial than it first appears with flexible quorums is what to do when we detect a higher ballot. If we detect a higher ballot that also `is_prepare_quorum()`, that's fine. But if we detect a higher ballot that, say, only `is_accept_quorum` but not `is_prepare_quorum`, it becomes less clear what to do. That server may or may not have completed the prepare phase. If it has completed the prepare phase, then it is correct to treat it as the leader. If not, then we will incorrectly continue to follow this stale leader.

I can think of two options on how to solve this, each with their own pros/cons:

1. Do an `is_prepare_quorum` check before we start following a higher ballot. This is the safest option but might cause unnecessary leader changes e.g., a leader was in accept phase and committing entries and then reconnects to another server that unnecessarily causes a leader change because the leader does not pass `is_prepare_quorum` check.
2. Incorporate more metadata in the ballots, such as if the servers are in the Prepare or Accept phase. This can probably avoid the unnecessary leader changes in (1) but starts to intermix BLE and Sequence Paxos logic.

At this point, I would prefer (1). The second option needs more investigation and proof of correctness before I feel comfortable to possibly merge it.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -32,32 +34,31 @@ impl Ballot {
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct BLEBallot {
     /// Ballot of a replica.
-    ballot: Ballot,
-    /// Custom priority parameter
-    priority: u32,
+    pub ballot: Ballot,
     /// Used to determine if the replica is a candidate to become a leader or remain a leader.
-    connectivity: u8,
+    pub connectivity: u8,
```
haraldng: This struct is nice for code readability i.e. having ballots: `Vec<BLEBallot>` instead of `Vec<(Ballot, u8)`. But I prefer not to have it as a public struct as that might confuse users (with `Ballot)` in the API docs. I think we should have it as a private struct and send `Ballot` and `connectivity` as separate fields in `HeartbeatReply`. We can then construct the BLEBallot before we add it to `self.ballots`

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -99,16 +100,11 @@ pub(crate) struct BallotLeaderElection {
     /// Vector which holds all the received heartbeats
     ballots: Vec<BLEBallot>,
     /// Holds the current ballot of this instance.
-    current_ballot: Ballot, // (round, pid)
-    /// The number of replicas inside the cluster that this instance is connected to (based on
-    /// heartbeats received) including itself.
-    connectivity: u8,
+    current_ballot: BLEBallot,
```
haraldng: As a result of the previous comment regarding making `BLEBallot` private, it probably is better to bring back `current_ballot: Ballot` and `connectivity: u8 ` as fields in BLE. I also believe that it will make the code for BLE more understandable, since we can easier see what fields are in BLE rather than them being hidden behind `BLEBallot`.

haraldng (changes requested):
Looks good. Comments are mainly related to code readability and clarity for users. Sorry for having you revert some changes I previously requested.

omnipaxos/src/messages.rs
```diff
@@ -224,7 +227,9 @@ pub mod ballot_leader_election {
         /// Number of the current heartbeat round.
         pub round: u32,
         /// BLEBallot of a replica.
-        pub ballot: BLEBallot,
+        pub ballot: Ballot,
+        /// The number of replicas inside the cluster the sender is connected to (including itself)
+        pub connectivity: Connectivity,
```
haraldng, line 232: I think the connectivity alias is mainly to make the BLE code more readable so I prefer not to expose it publicly.
```suggestion
        pub connectivity: u8,
```

omnipaxos/src/util.rs
```diff
@@ -278,6 +278,8 @@ pub type TrimmedIndex = u64;
 pub type NodeId = u64;
 /// ID for an OmniPaxos configuration (i.e., the set of servers in an OmniPaxos cluster)
 pub type ConfigurationId = u32;
+/// The connectivity of an OmniPaxos node
+pub type Connectivity = u8;
```
haraldng, line 282: Since we don't want to make this public and we only used it in BLE, I suggest moving this (without `pub`) to `ballot_leader_election.rs`

haraldng (changes requested):
Make the `Connectivity` alias private, and it should be ready.

omnipaxos/src/messages.rs
haraldng: Yes

haraldng (commented):
Looks good, I think updating the /docs with examples of how to use flexible quorums is missing. Though, might be easier to write the code examples after rebasing.

docs/introduction.md
```diff
@@ -1,5 +1,5 @@
 
-The OmniPaxos library is mainly driven by the [**OmniPaxos**](omnipaxos/index.md) struct.  It is a plain Rust struct and the user therefore needs to provide a network implementation themselves to actually send and receive messages. In this tutorial we will show how a user should interact with this struct in order to implement a strongly consistent, replicated log. This tutorial will focus on how to use the library and showcase its features. 
+The OmniPaxos library is mainly driven by the [**OmniPaxos**](omnipaxos/index.md) struct. It is a plain Rust struct and the user therefore needs to provide a network implementation themselves to actually send and receive messages. In this tutorial we will show how a user should interact with this struct in order to implement a strongly consistent, replicated log. This tutorial will focus on how to use the library and showcase its features.
```
haraldng, line 2: Cross references are a bit special in our case because it gets rendered to the website differently. We need to remove the `.md` suffix. 

More info here: https://github.com/haraldng/omnipaxos/pull/95#discussion_r1200877438

docs/omnipaxos/leader_election.md
haraldng, line 1: I think flexible quorums should be its own file/page. Can also make the first mention of "flexible quorums" clickable with a link to the paper.

docs/omnipaxos/leader_election.md
```diff
@@ -10,3 +10,29 @@ omni_paxos.election_timeout();
...
+    configuration_id: 1,
+    pid: 1,
+    peers: vec![2, 3, 4, 5, 6, 7],
+    flexible_quorum: Some(flex_quorum),
+    ..Default::deafult()
+};
+```
+In OmniPaxos, becoming a leader involves reading from the log and so `read_quorum_size` is the number of nodes to consult in order to become a leader. The `write_quorum_size` is the number of nodes to consult in order to append to the log.
+
+Appending to the log is a much more common occurance than electing a leader so reducing the `write_quorum_size` can reduce the latency and sometimes even increase throughput. Note, however, that reducing the `write_quorum_size` also decreases the fault tolerance of the cluster to `write_quorum_size - 1`.
```
haraldng, line 36:
```suggestion
Appending to the log is a much more common occurrence than electing a leader, so reducing the `write_quorum_size` can be beneficial. Note, however, that reducing the `write_quorum_size` also requires a larger `read_quorum_size`, which may reduce fault tolerance.
```

haraldng: I saw that you wrote something similar right afterwards. Then I would consider removing the last sentence here.

haraldng (changes requested):

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -60,24 +60,22 @@ where
...
+                        from: pid,
+                        to: *peer_pid,
+                        msg: PaxosMsg::PrepareReq,
+                    });
+                }
+                (state, b, None)
             }
             None => {
-                match &config.skip_prepare_use_leader {
+                match &config.initial_leader {
```
haraldng, line 78: The more I see this and other parts of the code related to this, I'm leaning toward removing this feature/field in the config. I see several reasons to remove it:

1. It makes initialisation and the failure-recovery messy with promised ballot vs. initial leader.
2. It makes the sequence numbers messy with a special case first accept case.

What do you think @kevin-harrison? I would propose to remove it, since the benefit of it is quite minimal i.e., just one skipped prepare phase (which is probably very light-weight anyway since there are no entries), but the code gets much more messy.

omnipaxos/src/util.rs
```diff
@@ -337,7 +337,7 @@ impl SequenceNumber {
     pub(crate) fn check_msg_status(&self, msg_seq_num: SequenceNumber) -> MessageStatus {
         if msg_seq_num == SequenceNumber::PREDEFINED_LEADER_FIRST_ACCEPT {
             MessageStatus::First
-        } else if msg_seq_num.session < self.session {
+        } else if msg_seq_num <= *self {
```
haraldng, line 340: I guess this is correct, but it looks a bit strange that we ignore a message even if it is in the right session. I think this is fine if we move the `else if` of the `expected` case above this just for clearness.

haraldng (changes requested):

omnipaxos/src/sequence_paxos/mod.rs
kevin-harrison: I definitely agree, it will make things a lot simpler. Also now that I think about it  sequence numbers mess up recovering our previous leader after storage. This is because we don't also recover the state of our sequence numbers. So as a follower even if we recover with the correct leader we will almost always falsely detect a sequence break. As a leader if we recover our ballot and are still the leader our `leader_state.follower_sequence_nums` get reset causing our followers to falsely detect sequence breaks but more importantly we will be repeating sequence numbers which is unacceptable.

omnipaxos/src/util.rs
kevin-harrison: This was done to protect against duplicated messages. So even in the same session there are messages to ignore. One example is when we resend `AcceptStopSign` it has the same sequence number but we don't want to trigger a `PrepareReq`.

omnipaxos/src/sequence_paxos/mod.rs
haraldng: Not sure if I understand what you mean. But if a leader recovers after a crash, the idea is not for it to continue to lead in the same ballot (before it crashed). Instead, the BLE will elect a new higher ballot.

Isn't it that when a server recovers from a crash, it will go into the `Recover` mode which will reset the sequence numbers anyway (which is fine)?

omnipaxos/src/util.rs
haraldng: I understand that, it's just a bit strange to read the code and see this if clause before the next one (the expected case) which is `else if msg_seq_num.session == self.session && ...`. That's why I think you should put the `else if` for the `Expected` case above this one.

omnipaxos/src/sequence_paxos/mod.rs
kevin-harrison: What is the purpose of recovering ballots after a crash? Is it to skip a prepare phase or for safety?

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -410,4 +322,20 @@ where
         }
         Ok(())
     }
+
+    /// Also returns if the incoming sequence number is the expected one.
+    fn handle_sequence_num(&mut self, seq_num: SequenceNumber, from: NodeId) -> bool {
```
haraldng, line 327: return the MessageStatus instead of bool

haraldng (changes requested):
Looks good. Please change the `handle_sequence_num` to return `MessageStatus` and double check in docs that we don't refer to ballots as the removed `BLEBallot`.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -224,88 +222,31 @@ where
     pub(crate) fn handle_acceptdecide(&mut self, acc: AcceptDecide<T>) {
         if self.internal_storage.get_promise() == acc.n
             && self.state == (Role::Follower, Phase::Accept)
+            && self.handle_sequence_num(acc.seq_num, acc.n.pid)
```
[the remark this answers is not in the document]
kevin-harrison: I was. Forgot to include it in the push.

haraldng (approved):

PR #94, opened 2023-05-16
Reconfiguration configs

### LINKED ISSUE #66: StopSign with OmniPaxosConfig for next instance
Currently, the `StopSign` has a few fields describing the new configuration: `config_id`, `nodes`, and `metadata`. The user is then supposed to take these fields and manually create a new `OmniPaxos` instance (or via `OmniPaxosConfig`). This should be automated such that when the `StopSign` is read or somehow "built", an `Option<OmniPaxosConfig>` is returned which the user can just build and use in the next configuration.

One approach is to introduce a new struct `ReconfigurationConfig` that has a function `fn build_for(pid: NodeID) -> OmniPaxos` so that users can take the given configuration and build a new OmniPaxos instance (with different Entry and Storage) for a specific server. This way we will also initialize OmniPaxos correctly.

kevin-harrison:

## Issues
Fix #66. Reconfigurations now done by taking a decided `StopSign`'s `next_config: ClusterConfig` field and calling `next_config.build_for()`.

## Breaking Changes
 - Changes the structure of `OmniPaxosConfig` (and therefore TOML config)
 - Changes the structure of `StopSign`
 - `ReconfigurationRequest` removed
 - `Omnipaxos::reconfigure()` now takes a `ClusterConfig` as an argument instead of `ReconfigurationReqest`
 
 ## TODO
 - [x] Check if `next_config: ClusterConfig` is valid before proposing a reconfiguration
 - [ ] Update OmniPaxos book


haraldng:
Good start! I would probably call it `ServerConfig` instead of `InstanceConfig` (instance can also mean one omnipaxos instance which is like one cluster). If possible it would be nice to also remove `peers` from `ServerConfig` since we already have `nodes` in `InstanceConfig`

omnipaxos/src/omni_paxos.rs
```diff
@@ -73,18 +69,77 @@ impl OmniPaxosConfig {
     }
 }
 
-impl Default for OmniPaxosConfig {
+/// Configuration for an `OmniPaxos` cluster.
+/// # Fields
+/// * `configuration_id`: The identifier for the cluster configuration that this Sequence Paxos replica is part of.
+/// * `nodes`: The nodes in the cluster i.e. the `pid`s of the other replicas in the configuration.
+/// * `initial_leader`: The initial leader of the cluster. Could be used in combination with reconfiguration to skip the prepare phase when switching to a new configuration.
```
haraldng, line 76: Since we call the other config `ServerConfig`, it is better if our docs here say "server" instead of "replica". And also "OmniPaxos" instead of "Sequence Paxos".

omnipaxos/src/omni_paxos.rs
```diff
@@ -73,18 +69,77 @@ impl OmniPaxosConfig {
...
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(any(feature = "serde", feature = "toml_config"), derive(Deserialize))]
+#[cfg_attr(feature = "toml_config", serde(default))]
+#[cfg_attr(feature = "serde", derive(Serialize))]
+pub struct ClusterConfig {
+    /// The identifier for the cluster configuration that this Sequence Paxos replica is part of.
+    pub configuration_id: u32,
+    /// The nodes in the cluster i.e. the `pid`s of the other replicas in the configuration.
+    pub nodes: Vec<u64>,
```
haraldng, line 85: Use `NodeId` instead of u64

```suggestion
    pub nodes: Vec<NodeId>,
```

omnipaxos/src/omni_paxos.rs
```diff
@@ -73,18 +69,77 @@ impl OmniPaxosConfig {
...
+
+impl Default for ClusterConfig {
+    fn default() -> Self {
+        Self {
+            configuration_id: 1,
+            nodes: Vec::new(),
+            initial_leader: None,
+        }
+    }
+}
```
haraldng, line 113: Do we need default? I would maybe prefer to not provide a default implementation since users are supposed to set this. If we need it to derive serde then let's it's better to derive default, where configuration_id = 0

omnipaxos/src/omni_paxos.rs
```diff
@@ -210,9 +265,11 @@ where
         self.seq_paxos.append(entry)
     }
 
-    /// Propose a reconfiguration. Returns error if already stopped or new configuration is empty.
-    pub fn reconfigure(&mut self, rc: ReconfigurationRequest) -> Result<(), ProposeErr<T>> {
-        self.seq_paxos.reconfigure(rc)
+    /// Propose a reconfiguration. Returns error if already stopped.
+    pub fn reconfigure(&mut self, new_configuration: ClusterConfig) -> Result<(), ProposeErr<T>> {
+        // TODO: doesn't check to make sure the new_config is valid. A user could propose a flawed
+        // ClusterConfig and if it gets decided be stuck with it. Is this acceptable?
```
haraldng, line 271: Perhaps we can extend `ProposeErr` with another variant `InvalidClusterConfig` that returns the ClusterConfig

omnipaxos/src/omni_paxos.rs
```diff
@@ -210,9 +265,11 @@ where
         self.seq_paxos.append(entry)
     }
 
-    /// Propose a reconfiguration. Returns error if already stopped or new configuration is empty.
-    pub fn reconfigure(&mut self, rc: ReconfigurationRequest) -> Result<(), ProposeErr<T>> {
-        self.seq_paxos.reconfigure(rc)
+    /// Propose a reconfiguration. Returns error if already stopped.
```
haraldng, line 268: Describe the parameter `new_configuration`

omnipaxos/src/omni_paxos.rs
```diff
@@ -73,18 +69,77 @@ impl OmniPaxosConfig {
...
+    pub configuration_id: u32,
+    /// The nodes in the cluster i.e. the `pid`s of the other replicas in the configuration.
+    pub nodes: Vec<u64>,
+    /// The initial leader of the cluster. Could be used in combination with reconfiguration to skip the prepare phase when switching to a new configuration.
+    pub initial_leader: Option<Ballot>,
+}
+
+impl ClusterConfig {
+    /// Checks all configurations and returns the local OmniPaxos node if successful.
+    pub fn build_for<T, B>(self, storage: B, server_config: ServerConfig) -> OmniPaxos<T, B>
```
haraldng, line 92:
```suggestion
    pub fn build_for_server<T, B>(self, config: ServerConfig, with_storage: B) -> OmniPaxos<T, B>
```

omnipaxos/src/omni_paxos.rs
```diff
@@ -73,18 +69,77 @@ impl OmniPaxosConfig {
...
+    /// The identifier for the cluster configuration that this Sequence Paxos replica is part of.
+    pub configuration_id: u32,
+    /// The nodes in the cluster i.e. the `pid`s of the other replicas in the configuration.
+    pub nodes: Vec<u64>,
+    /// The initial leader of the cluster. Could be used in combination with reconfiguration to skip the prepare phase when switching to a new configuration.
+    pub initial_leader: Option<Ballot>,
+}
+
+impl ClusterConfig {
+    /// Checks all configurations and returns the local OmniPaxos node if successful.
```
haraldng, line 91: add descriptions of parameters.

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -272,40 +274,36 @@ where
...
+            "Propose reconfiguration {:?}", new_config.nodes
         );
         if self.stopped() {
-            Err(ProposeErr::Reconfiguration(new_configuration))
+            Err(ProposeErr::Reconfiguration(new_config))
         } else {
             match self.state {
                 (Role::Leader, Phase::Prepare) => {
                     if self.pending_stopsign.is_none() {
-                        let ss = StopSign::with(self.config_id + 1, new_configuration, metadata);
```
haraldng, line 292: It would be nice to still be able to have some user-provided metadata that gets decided with the reconfiguration. Probably it is better to have it as a second argument for `reconfigure()`.

omnipaxos/tests/utils.rs
```diff
@@ -695,14 +692,9 @@ impl Entry for Value {
...
-    let v = ss
-        .metadata
-        .as_ref()
-        .expect("StopSign Metadata was None")
-        .first()
-        .expect("Empty metadata");
-    Value(*v as u64)
+fn stopsign_config_to_value(ss: &StopSign) -> Value {
+    let v = ss.next_config.configuration_id;
+    Value(v as u64)
```
haraldng, line 697: This looks a bit weird. I guess you wanted to use the ask-reply in kompact to check that the reconfiguration was decided. In that case, it is better to explicitly create the Value out of the configuration id (and comment that in the code) instead of having this function.

haraldng (changes requested):
Looks good. Please re-introduce the metadata (which can be useful for specifying version numbers or migration schemes) and fix other comments.

omnipaxos/src/omni_paxos.rs
kevin-harrison: I'm not a fan of the name `ServerConfig`, to me it implies the whole network layer as well. How about `ReplicaConfig`?

omnipaxos/src/omni_paxos.rs
kevin-harrison: The advantage of default is it allows users to more simply configure only the things they care about. Like how you've shown in the book before:
```rust
let omnipaxos_config = OmniPaxosConfig {
    configuration_id,
    pid: my_pid,
    peers: my_peers,
    ..Default::default()
}
```
Also it gives an idea to users about what a reasonable config is. The problem is that there really is no default `pid`, or `nodes`.

haraldng: Yes, that is precisely my point, there are no good default values for pid, peers, and configuration_id. We should therefore not provide a default implementation for it. When they code and get forced to specify these values (rather than just use default to make the compiler happy), they will look at the docs/examples to see how to set them, which is what we want.

Default for other configs make sense when we have many fields and there are good default values that can be set. This is not the case here in my opinion.

haraldng: But as I said, if we really need it for ser/de then keep it but make the default values fail to build.

omnipaxos/src/errors.rs
```diff
@@ -1,43 +1,66 @@
...
+//     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+//         match *self {
+//             #[cfg(feature = "toml_config")]
+//             ConfigError::ReadFile(ref err) => Some(err),
+//             #[cfg(feature = "toml_config")]
+//             ConfigError::Parse(ref err) => Some(err),
+//             ConfigError::InvalidConfig(ref str) => Some(self),
+//         }
+//     }
+// }
```
haraldng, line 42: remove if not needed

omnipaxos/src/omni_paxos.rs
```diff
@@ -132,6 +133,14 @@ pub struct ServerConfig {
     pub leader_priority: u64,
 }
 
+impl ServerConfig {
+    fn validate(&self) -> Result<(), ConfigError> {
+        valid_config!(self.pid != 0, "Initial leader pid cannot be 0");
```
haraldng, line 138:
```suggestion
        valid_config!(self.pid != 0, "Server pid cannot be 0");
```

omnipaxos/src/omni_paxos.rs
```diff
@@ -295,15 +311,13 @@ where
...
+/// An error indicating a failed proposal due to the current configuration being already stopped
+/// or due to an invalid proposed configuration.
 #[allow(missing_docs)]
 #[derive(Debug)]
-pub enum ProposeErr<T>
-where
-    T: Entry,
-{
-    Normal(T),
-    Reconfiguration(ClusterConfig),
```
haraldng: I think the naming/documentation was not good before. But a ProposeErr returns what was failed to be proposed, and I want users to be able to get back the `T` that failed to get proposed.

omnipaxos/src/omni_paxos.rs
```diff
@@ -295,15 +311,13 @@ where
...
+pub enum ProposeErr {
+    Stopped,
+    Reconfiguration(ConfigError),
 }
```
haraldng, line 321: I'm thinking that we should get rid of this enum because they are for two different things.

`Stopped` is for when calling `append() `while `Reconfiguration` is for calling `reconfigure()`.

I also suggest to rename `Stopped` to `PendingReconfiguration`, since it might not actually be stopped, but only pending (the StopSign could even later be dropped).

haraldng: Probably this is a result my bad naming also for the `stopped()` function. That should be renamed to `pending_reconfiguration()`

haraldng (changes requested):

haraldng (approved):

omnipaxos/src/omni_paxos.rs
```diff
@@ -24,87 +26,185 @@ use toml;
...
+    pub flexible_quorum: Option<FlexibleQuorum>,
+}
+
+impl ClusterConfig {
+    /// Checks that all the fields of the cluster config are valid
+    fn validate(&self) -> Result<(), ConfigError> {
+        let num_nodes = self.nodes.len();
+        valid_config!(num_nodes > 1, "Need more than 1 node");
+        valid_config!(self.configuration_id != 0, "Configuration ID cannot be 0");
+        if let Some(FlexibleQuorum {
```
kevin-harrison, line 107: I moved the FlexibleQuorum check back here. Before the validity of the flexible quorum was only checked when we constructed SequencePaxos, which meant a user could propose a SS with an invalid flexible quorum. The check cant be done when creating the FlexibleQuorum struct since we need to know the # of nodes.

omnipaxos/src/omni_paxos.rs
```diff
@@ -277,37 +391,17 @@ where
...
-            new_configuration,
-            metadata,
-        }
-    }
-}
-
-/// An error returning the proposal that was failed due to that the current configuration is stopped.
+/// An error indicating a failed proposal due to the current cluster configuration being already stopped
+/// or due to an invalid proposed configuration. Returns the failed proposal.
 #[allow(missing_docs)]
```
haraldng, line 396: Let's describe each variant since they are not very intuitive i.e., describe that those PendingReconfig variants are due to a pending reconfiguration and they return the failed entry.
```suggestion
```

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -27,9 +27,11 @@ where
     B: Storage<T>,
 {
     pub(crate) internal_storage: InternalStorage<B, T>,
+    // Will be needed after rebase
+    #[allow(dead_code)]
```
haraldng, line 31: remove?

omnipaxos/src/omni_paxos.rs
```diff
@@ -24,87 +26,185 @@ use toml;
...
+/// * `flexible_quorum` : Defines read and write quorum sizes. Can be used for different latency vs fault tolerance tradeoffs.
+#[derive(Clone, Debug, PartialEq, Default)]
+#[cfg_attr(any(feature = "serde", feature = "toml_config"), derive(Deserialize))]
+#[cfg_attr(feature = "toml_config", serde(default))]
+#[cfg_attr(feature = "serde", derive(Serialize))]
+pub struct ClusterConfig {
+    /// The identifier for the cluster configuration that this OmniPaxos server is part of. Must
+    /// not be 0 and be greater than the previous configuration's id.
+    pub configuration_id: ConfigurationId,
+    /// The nodes in the cluster i.e. the `pid`s of the other servers in the configuration.
```
haraldng, line 95:
```suggestion
    /// The nodes in the cluster i.e. the `pid`s of the servers in the configuration.
```

haraldng (changes requested):
Looks good, please fix requested small changes and update docs/comments.

omnipaxos/src/omni_paxos.rs
```diff
@@ -393,14 +393,18 @@ where
 
 /// An error indicating a failed proposal due to the current cluster configuration being already stopped
 /// or due to an invalid proposed configuration. Returns the failed proposal.
-#[allow(missing_docs)]
 #[derive(Debug)]
 pub enum ProposeErr<T>
 where
     T: Entry,
 {
+    /// Couldn't propose entry because cluster is stopped. Contains failed, proposed entry.
```
haraldng, line 401:
```suggestion
    /// Couldn't propose entry because a reconfiguration is pending. Returns the failed, proposed entry.
```

omnipaxos/src/omni_paxos.rs
```diff
@@ -393,14 +393,18 @@ where
...
     PendingReconfigEntry(T),
+    /// Couldn't propose reconfiguration because cluster is already stopped. Contains failed, proposed
```
haraldng, line 403:
```suggestion
    /// Couldn't propose reconfiguration because a reconfiguration is already pending. Returns the failed, proposed `ClusterConfig` and the metadata.
```

haraldng (approved):

PR #98, opened 2023-05-31
Refactor in favour of single source of truth for stopsign decidedness

### LINKED ISSUE #96: Implicitly decided stopsign during prepare phase
The `Promise` and `AcceptSync` messages both contain a stopsign and the decided index. However the stopsign is never treated as decided in these messages. Unfortunately the decided index could still imply that the stopsign should be decided and this would also be reflected in storage, if the message is handled without errors.

It is not clear to me if this is a critical bug or just a quirk of the implementation, but it should be fixed.

In my opinion, the real issue here is that the storage API defines 2 values that indicate if a stopsign is decided or not (decided index, stopsign entry). This means there is no single source of truth on whether a stopsign is decided. This should only be in one value that will be the ground truth on this. If the API should still reflect this information in both values, this should be generated from that ground truth, most likely in internal storage.

JonathanArns:

## Issues
fixes #96 

## Breaking Changes
The `StopSignEntry` which is also read by the public API, does not contain a `decided` flag anymore, but instead an `idx`.
I don't think this is very noteworthy, since only decided `StopSign`s are read anyways.

### DIFF
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -126,7 +126,7 @@ where
                 vec![RollbackValue::AcceptedRound(old_accepted_round)],
                 "storage error while trying to write decided index",
             );
-            let accepted = match accsync.decided_snapshot {
+            let mut accepted = match accsync.decided_snapshot {
                 Some(s) => {
                     let result = match s {
                         SnapshotType::Complete(c) => {
@@ -176,15 +176,6 @@ where
                     }
                 }
             };
-            self.state = (Role::Follower, Phase::Accept);
-            self.current_seq_num = accsync.seq_num;
-            let cached_idx = self.outgoing.len();
-            self.latest_accepted_meta = Some((accsync.n, cached_idx));
-            self.outgoing.push(PaxosMessage {
-                from: self.pid,
-                to: from,
-                msg: PaxosMsg::Accepted(accepted),
-            });
             match accsync.stopsign {
                 Some(ss) => {
                     if let Some(ss_entry) = self
@@ -198,15 +189,19 @@ where
                     } else {
                         self.accept_stopsign(ss);
                     }
-                    let a = AcceptedStopSign { n: accsync.n };
-                    self.outgoing.push(PaxosMessage {
-                        from: self.pid,
-                        to: from,
-                        msg: PaxosMsg::AcceptedStopSign(a),
-                    });
+                    accepted.accepted_idx += 1;
                 }
                 None => self.forward_pending_proposals(),
             }
+            self.state = (Role::Follower, Phase::Accept);
+            self.current_seq_num = accsync.seq_num;
+            let cached_idx = self.outgoing.len();
+            self.latest_accepted_meta = Some((accsync.n, cached_idx));
+            self.outgoing.push(PaxosMessage {
+                from: self.pid,
+                to: from,
+                msg: PaxosMsg::Accepted(accepted),
+            });
         }
     }
 
@@ -292,6 +287,10 @@ where
             && self.state == (Role::Follower, Phase::Accept)
         {
             let msg_status = self.current_seq_num.check_msg_status(acc_ss.seq_num);
+            let log_len = self
+                .internal_storage
+                .get_log_len()
+                .expect("storage error while trying to read log_len");
             match msg_status {
                 MessageStatus::First => {
                     // pseudo-AcceptSync for prepare-less reconfigurations
@@ -310,11 +309,14 @@ where
             }
 
             self.accept_stopsign(acc_ss.ss);
-            let a = AcceptedStopSign { n: acc_ss.n };
+            let a = Accepted {
+                n: acc_ss.n,
+                accepted_idx: log_len + 1,
+            };
             self.outgoing.push(PaxosMessage {
                 from: self.pid,
                 to: self.leader.pid,
-                msg: PaxosMsg::AcceptedStopSign(a),
+                msg: PaxosMsg::Accepted(a),
             });
         }
     }
@@ -346,51 +348,6 @@ where
             }
             self.internal_storage
                 .set_decided_idx(dec.decided_idx)
-                .expect("storage error while trying to write decided index");
-        }
-    }
-
-    pub(crate) fn handle_decide_stopsign(&mut self, dec: DecideStopSign) {
-        if self
-            .internal_storage
-            .get_promise()
-            .expect("storage error while trying to read promise")
-            == dec.n
-            && self.state.1 == Phase::Accept
-        {
-            let msg_status = self.current_seq_num.check_msg_status(dec.seq_num);
-            match msg_status {
-                MessageStatus::First => {
-                    #[cfg(feature = "logging")]
-                    warn!(
-                        self.logger,
-                        "DecideStopSign cannot be the first message in a sequence!"
-                    );
-                    return;
-                }
-                MessageStatus::Expected => self.current_seq_num = dec.seq_num,
-                MessageStatus::DroppedPreceding => {
-                    self.reconnected(dec.n.pid);
-                    return;
-                }
-                MessageStatus::Outdated => return,
-            }
-
-            let _ss = self
-                .internal_storage
-                .get_stopsign()
-                .expect("storage error while trying to read stopsign")
-                .expect("No stopsign found when deciding!");
-            let log_len = self
-                .internal_storage
-                .get_log_len()
-                .expect("storage error while trying to read log length");
-            debug_assert!(
-                _ss.idx == log_len,
-                "trying to decide StopSignEntry with wrong index"
-            );
-            self.internal_storage
-                .set_decided_idx(log_len + 1)
                 .expect("storage error while trying to write decided index");
         }
     }
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -192,11 +192,7 @@ where
                         .get_stopsign()
                         .expect("storage error while trying to read stopsign")
                     {
-                        let StopSignEntry {
-                            decided: has_decided,
-                            stopsign: _my_ss,
-                        } = ss_entry;
-                        if !has_decided {
+                        if !ss_entry.decided(old_decided_idx) {
                             self.accept_stopsign(ss);
                         }
                     } else {
@@ -380,26 +376,22 @@ where
                 MessageStatus::Outdated => return,
             }
 
-            let mut ss = self
+            let _ss = self
                 .internal_storage
                 .get_stopsign()
                 .expect("storage error while trying to read stopsign")
                 .expect("No stopsign found when deciding!");
-            ss.decided = true;
             let log_len = self
                 .internal_storage
                 .get_log_len()
                 .expect("storage error while trying to read log length");
-            let old_decided_idx = self.get_decided_idx();
+            debug_assert!(
+                _ss.idx == log_len,
+                "trying to decide StopSignEntry with wrong index"
+            );
             self.internal_storage
                 .set_decided_idx(log_len + 1)
                 .expect("storage error while trying to write decided index");
-            let result = self.internal_storage.set_stopsign(ss); // need to set it again now with the modified decided flag
-            self.internal_storage.rollback_if_err(
-                &result,
-                vec![RollbackValue::DecidedIdx(old_decided_idx)],
-                "storage error while trying to write decided index",
-            );
         }
     }
 
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -183,13 +183,7 @@ where
                     } else {
                         self.accept_stopsign(ss);
                     }
-                    let ss_idx = self
-                        .internal_storage
-                        .get_stopsign()
-                        .expect("stopsign should have been accepted")
-                        .log_idx;
-                    assert_eq!(ss_idx, accepted.accepted_idx);
-                    accepted.accepted_idx = ss_idx + 1;
+                    accepted.accepted_idx = self.internal_storage.get_accepted_idx();
                 }
                 None => self.forward_pending_proposals(),
             }
@@ -241,16 +235,9 @@ where
             && self.handle_sequence_num(acc_ss.seq_num, acc_ss.n.pid) == MessageStatus::Expected
         {
             self.accept_stopsign(acc_ss.ss);
-            let ss_idx = self
-                .internal_storage
-                .get_stopsign()
-                .expect("stopsign not accepted")
-                .log_idx;
-            let accepted_idx = self.internal_storage.get_accepted_idx();
-            assert_eq!(ss_idx + 1, accepted_idx);
             let a = Accepted {
                 n: acc_ss.n,
-                accepted_idx,
+                accepted_idx: self.internal_storage.get_accepted_idx(),
             };
             self.outgoing.push(PaxosMessage {
                 from: self.pid,
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -30,8 +30,8 @@ where
                 .expect("storage error while trying to read accepted round");
             let accepted_idx = self
                 .internal_storage
-                .get_log_len()
-                .expect("storage error while trying to read log length");
+                .get_accepted_idx()
+                .expect("storage error while trying to read accepted_idx");
             let decided_idx = self.get_decided_idx();
             let stopsign = self.get_stopsign();
             let (decided_snapshot, suffix) = if na > prep.n_accepted {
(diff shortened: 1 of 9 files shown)


JonathanArns:
I'd say we'll merge this last out of the ready PRs, since resolving the conflicts shouldn't be too hard here.

omnipaxos/src/storage.rs
```diff
@@ -28,13 +28,18 @@ pub trait Entry: Clone + Debug {
 #[allow(missing_docs)]
 pub struct StopSignEntry {
     pub stopsign: StopSign,
-    pub decided: bool,
+    pub idx: u64,
```
haraldng, line 31: just to be ultra clear:
```suggestion
    /// The log index of this StopSign
    pub log_idx: u64,
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -380,26 +376,22 @@ where
                 MessageStatus::Outdated => return,
             }
 
-            let mut ss = self
+            let _ss = self
                 .internal_storage
                 .get_stopsign()
                 .expect("storage error while trying to read stopsign")
                 .expect("No stopsign found when deciding!");
```
haraldng, line 383: I guess we can remove this (to avoid doing get_stopsign) unnecessarily?

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -592,30 +592,22 @@ where
         {
             self.leader_state.set_accepted_stopsign(from);
             if self.leader_state.is_stopsign_chosen() {
-                let mut ss = self
+                let _ss = self
                     .internal_storage
                     .get_stopsign()
                     .expect("storage error while trying to read stopsign")
                     .expect("No stopsign found when deciding!");
```
haraldng, line 599: Same here, remove

haraldng (changes requested):
Looks good, just a few small comments + docs.

I'm also thinking if it makes sense to remove the messages `AcceptedStopSign` and  `DecideStopSign`? Then the accepted-decided are just handled as normal entries.

JonathanArns:
I removed the `AcceptedStopSign` and `DecideStopSign` messages now, also changed the `LeaderState` accordingly.

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -237,9 +237,13 @@ where
...
                     {
-                        if !ss.decided {
+                        let decided_idx = self
+                            .internal_storage
+                            .get_decided_idx()
+                            .expect("storage error while trying to read decided_idx");
+                        if !ss.decided(decided_idx) {
                             for follower in self.leader_state.get_promised_followers() {
-                                if !self.leader_state.follower_has_accepted_stopsign(follower) {
+                                if !self.leader_state.get_accepted_idx(follower) > ss.log_idx {
```
haraldng, line 246: This comparison looks a bit weird, maybe just do
```suggestion
                                if self.leader_state.get_accepted_idx(follower) < ss.log_idx {
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -176,41 +176,32 @@ where
...
                     } else {
                         self.accept_stopsign(ss);
                     }
-                    let a = AcceptedStopSign { n: accsync.n };
-                    self.outgoing.push(PaxosMessage {
-                        from: self.pid,
-                        to: from,
-                        msg: PaxosMsg::AcceptedStopSign(a),
-                    });
+                    accepted.accepted_idx += 1;
```
haraldng, line 192: Here I would like to add an `assert_eq!(ss.accepted_index, accepted.accepted_idx +1)` (it's ok to add an assert here since we only handle acceptsync when leader changes). And then set `accepted.accepted_idx = ss.accepted_index`

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -314,11 +309,14 @@ where
             }
 
             self.accept_stopsign(acc_ss.ss);
-            let a = AcceptedStopSign { n: acc_ss.n };
+            let a = Accepted {
+                n: acc_ss.n,
+                accepted_idx: log_len + 1,
+            };
```
haraldng, line 315: Same here, add assert and reply with the accepted_index from the stopsign

haraldng (changes requested):
Looks good. Just need to add some more sanity checks and fix small comments.

haraldng:
As @kevin-harrison found out, we need to update the accepted index when accepting a stopsign so that the prepare phase is correct (i.e., a stopsign can get adopted). This probably requires to change how the accepted index is calculated in `InternalStorage`

kevin-harrison:
Right now we don't increment our accepted_idx in storage when we accept a stopsign (the StopSign PR dosen't seem to do this either). The accepted_idx becomes disconnected from the true state of the log. This causes problems, for example, when a leader wants to figure out the max promise of its followers to sync itself. If the followers are ahead by only a stopsign entry then the leader will see them self as equally up-to-date since the followers haven't incremented their accepted_idx .
Tim and I started changing my flexible-quorum merge branch to make the accepted_idx  include the stopsign, but this feature should probably be separated from that. Also I'm not sure how the changes interact with your new stopsign PR. Here is the unfinished work we started:
```diff
diff --git a/omnipaxos/src/sequence_paxos/follower.rs b/omnipaxos/src/sequence_paxos/follower.rs
index 366f3f6..21bf727 100644
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -190,7 +190,6 @@ where
                     if let Some(ss_entry) = self
                         .internal_storage
                         .get_stopsign()
-                        .expect("storage error while trying to read stopsign")
                     {
                         let StopSignEntry {
                             decided: has_decided,
@@ -364,13 +363,12 @@ where
             let mut ss = self
                 .internal_storage
                 .get_stopsign()
-                .expect("storage error while trying to read stopsign")
                 .expect("No stopsign found when deciding!");
             ss.decided = true;
             let accepted_idx = self.internal_storage.get_accepted_idx();
             let old_decided_idx = self.get_decided_idx();
             self.internal_storage
-                .set_decided_idx(accepted_idx + 1)
+                .set_decided_idx(accepted_idx)
                 .expect("storage error while trying to write decided index");
             let result = self.internal_storage.set_stopsign(ss); // need to set it again now with the modified decided flag
             self.internal_storage.rollback_and_panic_if_err(
diff --git a/omnipaxos/src/sequence_paxos/leader.rs b/omnipaxos/src/sequence_paxos/leader.rs
index 04fb44b..e88fb36 100644
--- a/omnipaxos/src/sequence_paxos/leader.rs
+++ b/omnipaxos/src/sequence_paxos/leader.rs
@@ -572,12 +572,11 @@ where
                 let mut ss = self
                     .internal_storage
                     .get_stopsign()
-                    .expect("storage error while trying to read stopsign")
                     .expect("No stopsign found when deciding!");
                 ss.decided = true;
                 let old_decided_idx = self.internal_storage.get_decided_idx();
                 self.internal_storage
-                    .set_decided_idx(self.internal_storage.get_accepted_idx() + 1)
+                    .set_decided_idx(self.internal_storage.get_accepted_idx())
                     .expect("storage error while trying to write decided index");
                 let result = self.internal_storage.set_stopsign(ss);
                 self.internal_storage.rollback_and_panic_if_err(
diff --git a/omnipaxos/src/sequence_paxos/mod.rs b/omnipaxos/src/sequence_paxos/mod.rs
index c3f9852..32cfb51 100644
--- a/omnipaxos/src/sequence_paxos/mod.rs
+++ b/omnipaxos/src/sequence_paxos/mod.rs
@@ -251,7 +251,6 @@ where
                     if let Some(ss) = self
                         .internal_storage
                         .get_stopsign()
-                        .expect("storage error while trying to read stopsign")
                     {
                         if !ss.decided {
                             for follower in self.leader_state.get_promised_followers() {
@@ -356,7 +355,6 @@ where
         match self
             .internal_storage
             .get_stopsign()
-            .expect("storage error while trying to read stopsign")
         {
             Some(ss) if ss.decided => Some(ss.stopsign),
             _ => None,
@@ -473,7 +471,6 @@ where
     fn get_stopsign(&self) -> Option<StopSign> {
         self.internal_storage
             .get_stopsign()
-            .expect("storage error while trying to read stopsign")
             .map(|x| x.stopsign)
     }
 }
diff --git a/omnipaxos/src/storage.rs b/omnipaxos/src/storage.rs
index 9228580..e1f255c 100644
--- a/omnipaxos/src/storage.rs
+++ b/omnipaxos/src/storage.rs
@@ -213,6 +213,8 @@ where
     compacted_idx: u64,
     /// Real length of logs in the storage.
     real_log_len: u64,
+    /// TODO: docs
+    stopsign: Option<StopSignEntry>,
 }
 
 impl<T> StateCache<T>
@@ -228,12 +230,17 @@ where
             decided_idx: 0,
             compacted_idx: 0,
             real_log_len: 0,
+            stopsign: None,
         }
     }
 
     // Returns the index of the last accepted entry.
     fn get_accepted_idx(&self) -> u64 {
-        self.compacted_idx + self.real_log_len
+        let mut accepted_idx = self.compacted_idx + self.real_log_len;
+        if let Some(_) = self.stopsign {
+            accepted_idx += 1;
+        }
+        accepted_idx
     }
 
     // Appends an entry to the end of the `batched_entries`. If the batch is full, the
@@ -374,7 +381,7 @@ where
         } else if idx < virtual_log_len {
             Ok(Some(IndexEntry::Entry))
         } else if idx == virtual_log_len {
-            match self.get_stopsign()? {
+            match self.get_stopsign() {
                 Some(ss) if ss.decided => Ok(Some(IndexEntry::StopSign(ss.stopsign))),
                 _ => Ok(None),
             }
@@ -388,7 +395,6 @@ where
     where
         R: RangeBounds<u64>,
     {
-        let accepted_idx = self.get_accepted_idx();
         let from_idx = match r.start_bound() {
             Bound::Included(i) => *i,
             Bound::Excluded(e) => *e + 1,
@@ -397,24 +403,19 @@ where
         let to_idx = match r.end_bound() {
             Bound::Included(i) => *i + 1,
             Bound::Excluded(e) => *e,
-            Bound::Unbounded => {
-                let idx = self.get_accepted_idx();
-                match self
-                    .get_stopsign()
-                    .expect("storage error while trying to read stopsign")
-                {
-                    Some(ss) if ss.decided => idx + 1,
-                    _ => idx,
-                }
-            }
+            Bound::Unbounded => self.get_accepted_idx(),
         };
+
+        // TODO: this needs to be based off of real_log_len not accepted_idx since that includes the
+        // stopsign now
         let compacted_idx = self.get_compacted_idx();
+        let real_log_len = self.state_cache.real_log_len;
         let to_type = match self.get_entry_type(to_idx - 1, compacted_idx, accepted_idx)? {
             // use to_idx-1 when getting the entry type as to_idx is exclusive
             Some(IndexEntry::Compacted) => {
                 return Ok(Some(vec![self.create_compacted_entry(compacted_idx)?]))
             }
-            Some(from_type) => from_type,
+            Some(to_type) => to_type,
             _ => return Ok(None),
         };
         let from_type = match self.get_entry_type(from_idx, compacted_idx, accepted_idx)? {
@@ -686,11 +687,13 @@ where
     }
 
     pub(crate) fn set_stopsign(&mut self, s: StopSignEntry) -> StorageResult<()> {
+        self.state_cache.stopsign = Some(s.clone());
         self.storage.set_stopsign(s)
     }
 
-    pub(crate) fn get_stopsign(&self) -> StorageResult<Option<StopSignEntry>> {
-        self.storage.get_stopsign()
+    pub(crate) fn get_stopsign(&self) -> Option<StopSignEntry> {
+        // TODO: is this clone necessary when we merge new StopSign PR?
+        self.state_cache.stopsign.clone()
     }
 
     pub(crate) fn create_snapshot(&mut self, compact_idx: u64) -> StorageResult<T::Snapshot> {
```

JonathanArns:
Okay so I applied the small feedback changes.
Then I merged the caching from master and then made the changes to `accepted_idx`.

omnipaxos/src/storage.rs
```diff
@@ -220,12 +231,17 @@ where
...
     }
 
     // Returns the index of the last accepted entry.
     fn get_accepted_idx(&self) -> u64 {
-        self.compacted_idx + self.real_log_len
+        let mut idx = self.compacted_idx + self.real_log_len;
+        if self.stopsign.is_some() {
+            idx += 1;
+        }
+        idx
```
haraldng, line 244:
```suggestion
        let idx = self.compacted_idx + self.real_log_len;
        if self.stopsign.is_some() {
            idx + 1
        } else {
            idx
        }    
```

omnipaxos/src/storage.rs
```diff
@@ -366,9 +382,12 @@ where
         } else if idx < virtual_log_len {
             Ok(Some(IndexEntry::Entry))
         } else if idx == virtual_log_len {
-            match self.get_stopsign()? {
-                Some(ss) if ss.decided => Ok(Some(IndexEntry::StopSign(ss.stopsign))),
-                _ => Ok(None),
+            match self.get_stopsign() {
+                Some(ss) => Ok(Some(IndexEntry::StopSign(ss.stopsign))),
+                _ => {
+                    println!("HELLLOOO");
```
haraldng, line 388:
```suggestion
```

omnipaxos/src/storage.rs
```diff
@@ -389,29 +407,27 @@ where
...
+        let log_len = self.state_cache.real_log_len + self.state_cache.compacted_idx;
+        let to_type = match self.get_entry_type(to_idx - 1, compacted_idx, log_len)? {
             // use to_idx-1 when getting the entry type as to_idx is exclusive
             Some(IndexEntry::Compacted) => {
                 return Ok(Some(vec![self.create_compacted_entry(compacted_idx)?]))
             }
             Some(from_type) => from_type,
-            _ => return Ok(None),
+            x => {
+                println!("TO {:?}", x);
```
haraldng, line 421:
```suggestion
```

omnipaxos/src/storage.rs
```diff
@@ -389,29 +407,27 @@ where
...
+                return Ok(None);
+            }
         };
-        let from_type = match self.get_entry_type(from_idx, compacted_idx, accepted_idx)? {
+        let from_type = match self.get_entry_type(from_idx, compacted_idx, log_len)? {
             Some(from_type) => from_type,
-            _ => return Ok(None),
+            x => {
+                println!("TO {:?}", x);
```
haraldng, line 428:
```suggestion
```

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -409,11 +402,13 @@ where
     }
 
     fn accept_stopsign(&mut self, ss: StopSign) {
+        let accepted_idx = self.internal_storage.get_accepted_idx();
```
haraldng, line 405:
```suggestion
        let ss_log_idx = self.internal_storage.get_accepted_idx();
```

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -409,11 +402,13 @@ where
...
         self.internal_storage
-            .set_stopsign(StopSignEntry::with(ss, false))
+            .set_stopsign(StopSignEntry::with(ss, accepted_idx))
             .expect("storage error while trying to write stopsign");
         if self.state.0 == Role::Leader {
-            self.leader_state.set_accepted_stopsign(self.pid);
+            self.leader_state
+                .set_accepted_idx(self.pid, accepted_idx + 1);
```
haraldng, line 411: Let's unify the access to the `accepted_idx`  by always calling `self.internal_storage.get_accepted_idx()`
```suggestion
            let accepted_idx = self.internal_storage.get_accepted_idx();
            assert_eq!(accepted_idx, ss_log_idx + 1);
            self.leader_state
                .set_accepted_idx(self.pid, accepted_idx);
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -248,11 +241,21 @@ where
...
         {
             self.accept_stopsign(acc_ss.ss);
-            let a = AcceptedStopSign { n: acc_ss.n };
+            let ss_idx = self
+                .internal_storage
+                .get_stopsign()
+                .expect("stopsign not accepted")
+                .log_idx;
+            let accepted_idx = self.internal_storage.get_accepted_idx();
+            assert_eq!(ss_idx + 1, accepted_idx);
```
haraldng, line 250: Move check to `accept_stopsign()` function (see prev comment)

```suggestion
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -174,41 +174,34 @@ where
...
-                        from: self.pid,
-                        to: from,
-                        msg: PaxosMsg::AcceptedStopSign(a),
-                    });
+                    let ss_idx = self
+                        .internal_storage
+                        .get_stopsign()
+                        .expect("stopsign should have been accepted")
+                        .log_idx;
+                    assert_eq!(ss_idx, accepted.accepted_idx);
```
haraldng, line 191: I think this check makes more sense to have inside the `accept_stopsign()` function (see comment there).

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -174,41 +174,34 @@ where
...
+                    accepted.accepted_idx = ss_idx + 1;
```
haraldng, line 192: Unify the access to accepted index by always calling the `get_accepted_idx()`

```suggestion
                    accepted.accepted_idx = self.internal_storage.get_accepted_idx();
```

haraldng (changes requested):
Please fix minor comments: avoid hardcoding +1 and use the `self.internal_storage.get_accepted_idx()`, move assertion check to `accept_stopsign` and remove printouts.

haraldng (approved):

PR #100, opened 2023-06-01
Atomic storage API rework

### LINKED ISSUE #97: Proposal: support truly atomic storage implementations
This is a proposal for a big breaking change to the storage API.

Even after merging the most recent atomic storage PR, my biggest gripe with the current implementation of Omnipaxos is the storage API and its lack of support for atomic transactions. Because the current API only allows writing individual values one at a time, we cannot really change multiple values atomically, even if the underlying storage implementation could support it.

The rollbacks we do now are a nice addition on top of the current API, but they only deal with individual errors from within storage. If an error occurs somewhere else and the application crashes for example, these rollbacks cannot help us.

I think we need a big change to the writing part of the storage API. My proposal is to replace all write functions on the API with one single function that takes a set of changes and atomically applies them. This could look something like this:
```rust
pub enum StorageUpdate {
    DecidedIdx(u64),
    AppendOnPrefix(Vec<Entry>, u64),
    ...
}

pub trait storage {
    // reading functions are unchanged
    ...
    /// Should perform all of the updates as an atomic set of changes
    fn update(Vec<StorageUpdate>) -> StorageResult<()>;
    ...
}
```
Implementing this API would not be significantly more complicated than the current one (provided one is familiar with enums and pattern matching in rust). It would easily support an implementation very similar to the current ones, where values are individually written to the underlying storage mechanism, but it does not enforce this anymore. Instead the storage implementation could also make use of the transaction mechanism of an embedded DB, to build a much more robust system than is currently possible.

Another plus is that we could remove all of the rollback code in omnipaxos, because this responsibility will be put on the storage implementation. (Omnipaxos is really complicated enough without this logic as well)

JonathanArns:

- [ ] You ran the local CI checker with `./check.sh` with no errors
- [ ] You reference which issue is being closed in the PR text (if applicable)

## Issues
fixes #97 

## Breaking Changes
This is supposed to be a major rework of the storage API (and everything that entails, likely touching a lot of places in the code), with the goal of enabling truly atomic storage implementations. The idea is described in the above issue in more detail.

We will try to stay _almost_ compatible with existing storage implementations, by providing a sort of compatibility layer that can provide basic atomicity for non-atomic storage implementations.

### DIFF
--- a/omnipaxos/src/storage.rs
+++ b/omnipaxos/src/storage.rs
@@ -23,6 +23,194 @@ pub trait Entry: Clone + Debug {
     type Snapshot: Snapshot<Self> + Serialize + for<'a> Deserialize<'a>;
 }
 
+
+///
+pub enum StorageOp<T: Entry> {
+    ///
+    AppendEntries{
+        ///
+        entries: Vec<T>,
+        ///
+        from_idx: Option<u64>
+    },
+    ///
+    SetPromise(Ballot),
+    ///
+    SetDecidedIndex(u64),
+    ///
+    SetAcceptedRound(Ballot),
+    ///
+    SetCompactedIdx(u64),
+    ///
+    Trim(u64),
+    ///
+    SetStopsign(Option<StopSignEntry>),
+    ///
+    SetSnapshot(Option<T::Snapshot>),
+}
+
+///
+pub trait AtomicStorage<T: Entry> {
+    /// Atomically perform and persist all ops in order.
+    /// Returns an Error if the transaction was not committed (so it was rolled back).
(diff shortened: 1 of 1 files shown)


JonathanArns:
So far all that has been done here is that @kevin-harrison and I drafted how the storage related types might look in the future.
At least `NonAtomicStorage` will likely need to change a little more, in order to hopefully accommodate rollbacks of log operations (append, trim). We should of course try to stay within the bounds of what is reasonable to implement in the context of an append log (so we don't enforce super-slow implementations in some way).

haraldng:
Superseded by #121.

PR #101, opened 2023-06-07
Feature resend decide ss

### LINKED ISSUE #68: Complete reconnection automation

kevin-harrison:

## Issues

Fix #68  by resending the decides of StopSigns.

## Breaking Changes
 - The trait function `Storage::set_stopsign(&mut self, s: StopSign) -> StorageResult<()>`'s signature changes to `Storage::set_stopsign(&mut self, s: Option<StopSign>) -> StorageResult<()>`.

## Other Changes
 - Fixed AcceptSync bugs by changing how Stopsign's and Snapshots are created.
 - Accounts for a bug in CommitLog when appending an empty set of entries.

### DIFF
--- a/docs/omnipaxos/reconfiguration.md
+++ b/docs/omnipaxos/reconfiguration.md
@@ -21,7 +21,7 @@ let decided_entries: Option<Vec<LogEntry<KeyValue>>> = omnipaxos.  read_decided_
 if let Some(de) = decided_entries {
     for d in de {
         match d {
-            LogEntry::StopSign(stopsign) => {
+            LogEntry::StopSign(stopsign, _stopsign_is_decided) => {
                 let new_configuration = stopsign.next_config;
                 if new_configuration.nodes.contains(&my_pid) {
                 // current configuration has been safely stopped. Start new instance
--- a/docs/omnipaxos/reconfiguration.md
+++ b/docs/omnipaxos/reconfiguration.md
@@ -21,7 +21,7 @@ let decided_entries: Option<Vec<LogEntry<KeyValue>>> = omnipaxos.  read_decided_
 if let Some(de) = decided_entries {
     for d in de {
         match d {
-            LogEntry::StopSign(stopsign, _stopsign_is_decided) => {
+            LogEntry::StopSign(stopsign, true) => {
                 let new_configuration = stopsign.next_config;
                 if new_configuration.nodes.contains(&my_pid) {
                 // current configuration has been safely stopped. Start new instance
--- a/omnipaxos/src/messages.rs
+++ b/omnipaxos/src/messages.rs
@@ -73,6 +73,8 @@ pub mod sequence_paxos {
         pub sync_idx: u64,
         /// The decided index
         pub decided_idx: u64,
+        /// The accepted index
+        pub accepted_idx: u64,
         /// StopSign to be accepted
         pub stopsign: Option<StopSign>,
     }
--- a/omnipaxos/src/messages.rs
+++ b/omnipaxos/src/messages.rs
@@ -73,8 +73,8 @@ pub mod sequence_paxos {
         pub sync_idx: u64,
         /// The decided index
         pub decided_idx: u64,
-        /// The accepted index
-        pub accepted_idx: u64,
+        /// The compacted index of the `decided_snapshot`
+        pub compacted_idx: u64,
         /// StopSign to be accepted
         pub stopsign: Option<StopSign>,
     }
--- a/omnipaxos/src/messages.rs
+++ b/omnipaxos/src/messages.rs
@@ -69,12 +69,10 @@ pub mod sequence_paxos {
         pub decided_snapshot: Option<SnapshotType<T>>,
         /// The log suffix.
         pub suffix: Vec<T>,
-        /// The index of the log where the entries from `sync_item` should be applied at or the compacted idx
+        /// The index of the log where the entries from `suffix` should be applied at (also the compacted idx of `decided_snapshot` if it exists)
         pub sync_idx: u64,
         /// The decided index
         pub decided_idx: u64,
-        /// The compacted index of the `decided_snapshot`
-        pub compacted_idx: u64,
         /// StopSign to be accepted
         pub stopsign: Option<StopSign>,
     }
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -37,7 +37,7 @@ where
                         .internal_storage
                         .get_suffix(decided_idx)
                         .expect("storage error while trying to read log suffix");
-                    (Some(delta_snapshot), suffix)
+                    (delta_snapshot, suffix)
                 } else {
                     let suffix = self
                         .internal_storage
@@ -56,7 +56,7 @@ where
                         .internal_storage
                         .get_suffix(decided_idx)
                         .expect("storage error while trying to read log suffix");
-                    (Some(delta_snapshot), suffix)
+                    (delta_snapshot, suffix)
                 } else {
                     let suffix = self
                         .internal_storage
@@ -123,9 +123,9 @@ where
                         SnapshotType::Complete(c) => {
                             self.internal_storage.set_snapshot(accsync.compacted_idx, c)
                         }
-                        SnapshotType::Delta(d) => {
-                            self.internal_storage.merge_snapshot(accsync.compacted_idx, d)
-                        }
+                        SnapshotType::Delta(d) => self
+                            .internal_storage
+                            .merge_snapshot(accsync.compacted_idx, d),
                     };
                     self.internal_storage.rollback_and_panic_if_err(
                         &snapshot_res,
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -25,7 +25,7 @@ where
             let na = self.internal_storage.get_accepted_round();
             let accepted_idx = self.internal_storage.get_accepted_idx();
             let decided_idx = self.get_decided_idx();
-            let stopsign = self.get_stopsign();
+            let stopsign = self.internal_storage.get_stopsign();
             let (decided_snapshot, suffix) = if na > prep.n_accepted {
                 let ld = prep.decided_idx;
                 if ld < decided_idx && T::Snapshot::use_snapshots() {
@@ -105,7 +105,7 @@ where
                 vec![RollbackValue::AcceptedRound(old_accepted_round)],
                 "storage error while trying to write decided index",
             );
-            let mut accepted = match accsync.decided_snapshot {
+            match accsync.decided_snapshot {
                 Some(s) => {
                     let old_compacted_idx = self.internal_storage.get_compacted_idx();
                     let old_log_res = self.internal_storage.get_suffix(old_compacted_idx);
@@ -119,12 +119,19 @@ where
                             "storage error while trying to read old log or snapshot",
                         );
                     }
+                    // Compact index of snapshot depends on if incoming stopsign is decided
+                    let new_compact_idx = match accsync.stopsign {
+                        Some(_) if accsync.decided_idx == accsync.accepted_idx => {
+                            accsync.decided_idx - 1
+                        }
+                        _ => accsync.decided_idx,
+                    };
                     let snapshot_res = match s {
                         SnapshotType::Complete(c) => {
-                            self.internal_storage.set_snapshot(accsync.decided_idx, c)
+                            self.internal_storage.set_snapshot(new_compact_idx, c)
                         }
                         SnapshotType::Delta(d) => {
-                            self.internal_storage.merge_snapshot(accsync.decided_idx, d)
+                            self.internal_storage.merge_snapshot(new_compact_idx, d)
                         }
                     };
                     self.internal_storage.rollback_and_panic_if_err(
@@ -148,11 +155,6 @@ where
                         ],
                         "storage error while trying to write log entries",
                     );
-                    Accepted {
-                        n: accsync.n,
-                        accepted_idx: accepted_res
-                            .expect("storage error while trying to write log entries"),
-                    }
                 }
                 None => {
                     // no snapshot, only suffix
@@ -167,26 +169,19 @@ where
                         ],
                         "storage error while trying to write log entries",
                     );
-                    Accepted {
-                        n: accsync.n,
-                        accepted_idx: accepted_idx
-                            .expect("storage error while trying to write log entries"),
-                    }
                 }
             };
-            match accsync.stopsign {
-                Some(ss) => {
-                    if let Some(ss_entry) = self.internal_storage.get_stopsign() {
-                        if !ss_entry.decided(old_decided_idx) {
-                            self.accept_stopsign(ss);
-                        }
-                    } else {
-                        self.accept_stopsign(ss);
-                    }
-                    accepted.accepted_idx = self.internal_storage.get_accepted_idx();
-                }
-                None => self.forward_pending_proposals(),
+            if accsync.stopsign.is_none() {
+                self.forward_pending_proposals();
+                //TODO: foward pending stopsign like in handle_majority_promises?
             }
+            self.internal_storage
+                .set_stopsign(accsync.stopsign)
+                .expect("storage error while trying to write stopsign");
+            let accepted = Accepted {
+                n: accsync.n,
+                accepted_idx: self.internal_storage.get_accepted_idx(),
+            };
             self.state = (Role::Follower, Phase::Accept);
             self.current_seq_num = accsync.seq_num;
             let cached_idx = self.outgoing.len();
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -31,7 +31,7 @@ where
                 if ld < decided_idx && T::Snapshot::use_snapshots() {
                     let delta_snapshot = self
                         .internal_storage
-                        .create_diff_snapshot(ld, decided_idx)
+                        .create_diff_snapshot(ld)
                         .expect("storage error while trying to read diff snapshot");
                     let suffix = self
                         .internal_storage
@@ -50,7 +50,7 @@ where
                 if T::Snapshot::use_snapshots() && compacted_idx > prep.accepted_idx {
                     let delta_snapshot = self
                         .internal_storage
-                        .create_diff_snapshot(prep.decided_idx, decided_idx)
+                        .create_diff_snapshot(prep.decided_idx)
                         .expect("storage error while trying to read diff snapshot");
                     let suffix = self
                         .internal_storage
@@ -119,19 +119,12 @@ where
                             "storage error while trying to read old log or snapshot",
                         );
                     }
-                    // Compact index of snapshot depends on if incoming stopsign is decided
-                    let new_compact_idx = match accsync.stopsign {
-                        Some(_) if accsync.decided_idx == accsync.accepted_idx => {
-                            accsync.decided_idx - 1
-                        }
-                        _ => accsync.decided_idx,
-                    };
                     let snapshot_res = match s {
                         SnapshotType::Complete(c) => {
-                            self.internal_storage.set_snapshot(new_compact_idx, c)
+                            self.internal_storage.set_snapshot(accsync.compacted_idx, c)
                         }
                         SnapshotType::Delta(d) => {
-                            self.internal_storage.merge_snapshot(new_compact_idx, d)
+                            self.internal_storage.merge_snapshot(accsync.compacted_idx, d)
                         }
                     };
                     self.internal_storage.rollback_and_panic_if_err(
@@ -173,7 +166,6 @@ where
             };
             if accsync.stopsign.is_none() {
                 self.forward_pending_proposals();
-                //TODO: foward pending stopsign like in handle_majority_promises?
             }
             self.internal_storage
                 .set_stopsign(accsync.stopsign)
--- a/omnipaxos/src/sequence_paxos/follower.rs
+++ b/omnipaxos/src/sequence_paxos/follower.rs
@@ -29,7 +29,7 @@ where
             let (decided_snapshot, suffix) = if na > prep.n_accepted {
                 let ld = prep.decided_idx;
                 if ld < decided_idx && T::Snapshot::use_snapshots() {
-                    let delta_snapshot = self
+                    let (delta_snapshot, _) = self
                         .internal_storage
                         .create_diff_snapshot(ld)
                         .expect("storage error while trying to read diff snapshot");
@@ -48,7 +48,7 @@ where
             } else if na == prep.n_accepted && accepted_idx > prep.accepted_idx {
                 let compacted_idx = self.internal_storage.get_compacted_idx();
                 if T::Snapshot::use_snapshots() && compacted_idx > prep.accepted_idx {
-                    let delta_snapshot = self
+                    let (delta_snapshot, _) = self
                         .internal_storage
                         .create_diff_snapshot(prep.decided_idx)
                         .expect("storage error while trying to read diff snapshot");
@@ -121,11 +121,11 @@ where
                     }
                     let snapshot_res = match s {
                         SnapshotType::Complete(c) => {
-                            self.internal_storage.set_snapshot(accsync.compacted_idx, c)
+                            self.internal_storage.set_snapshot(accsync.sync_idx, c)
                         }
-                        SnapshotType::Delta(d) => self
-                            .internal_storage
-                            .merge_snapshot(accsync.compacted_idx, d),
+                        SnapshotType::Delta(d) => {
+                            self.internal_storage.merge_snapshot(accsync.sync_idx, d)
+                        }
                     };
                     self.internal_storage.rollback_and_panic_if_err(
                         &snapshot_res,
@@ -137,7 +137,7 @@ where
                     );
                     let accepted_res = self
                         .internal_storage
-                        .append_entries_without_batching(accsync.suffix);
+                        .append_on_prefix(accsync.sync_idx, accsync.suffix);
                     self.internal_storage.rollback_and_panic_if_err(
                         &accepted_res,
                         vec![
--- a/omnipaxos/src/sequence_paxos/leader.rs
+++ b/omnipaxos/src/sequence_paxos/leader.rs
@@ -217,9 +217,11 @@ where
             .leader_state
             .get_decided_idx(*pid)
             .expect("Received PromiseMetaData but not found in ld");
-        let (delta_snapshot, suffix, sync_idx) =
-            if followers_decided_idx < my_decided_idx && T::Snapshot::use_snapshots() {
-                // Send missing decided entries in snapshot and accepted entries in suffix
+        let (delta_snapshot, suffix, sync_idx) = if followers_promise_n == max_promise_n {
+            // Follower's accepted entries are valid, send any new entries after
+            if T::Snapshot::use_snapshots() && *followers_accepted_idx < my_decided_idx {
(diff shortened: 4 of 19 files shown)


haraldng (commented):
I think the code looks mostly correct (please check comments to see if suggested changes make sense). 

I just want to understand the changes related to the snapshot better: How/when do we create an empty snapshot? and how does it break the decided index?

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -214,20 +214,21 @@ where
...
-                        if !ss.decided(decided_idx) {
-                            for follower in self.leader_state.get_promised_followers() {
-                                // resend if the follower has not accepted the stopsign
-                                if !ss.decided(self.leader_state.get_accepted_idx(follower)) {
-                                    self.send_accept_stopsign(follower, ss.stopsign.clone(), true);
-                                }
+
+                        for follower in self.leader_state.get_promised_followers() {
+                            if !ss.decided(decided_idx)
+                                && !ss.decided(self.leader_state.get_accepted_idx(follower))
```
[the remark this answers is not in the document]
kevin-harrison: Not 100% sure what you mean by losing last N accepts. But i think either the next message received is then either a prepare (we will resync and get missed accepts) or an acceptDecide/decide/acceptStopSign and the sequence break will be discovered.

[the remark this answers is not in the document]
kevin-harrison: When we create an empty stopsign:
- node A disconnects from cluster
- cluster proposes and decides a SS
- node reconnects to cluster
- leader creates AccSync with suffix=[ ], decided_idx=1 a stopsign and because the leader sees the it has a higher decided idx, it creates a snapshot of the empty log (can't snapshot a SS).
- on receiving the AccSync now Node A has a messed up log because it takes the SS and also the snapshot representing and empty log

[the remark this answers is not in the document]
haraldng: > Not 100% sure what you mean by losing last N accepts. But i think either the next message received is then either a prepare (we will resync and get missed accepts) or an acceptDecide/decide/acceptStopSign (the sequence break will be discovered).

Consider we send N Accept messages and then send a StopSign and they all get dropped. Since a StopSign is pending, the leader will not propose any further entries. Therefore we will never decide any of these unless we periodically send some message so that we can detect something has been dropped.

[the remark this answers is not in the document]
haraldng: but maybe we can solve that in a future PR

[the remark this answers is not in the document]
haraldng: on second thought, this is fine with your changes of resending stopsigns. But let's say we don't propose SS and don't propose any new entries after those N entries. Then those N entries will not be decided

omnipaxos/src/messages.rs
```diff
@@ -73,6 +73,8 @@ pub mod sequence_paxos {
         pub sync_idx: u64,
         /// The decided index
         pub decided_idx: u64,
+        /// The accepted index
+        pub accepted_idx: u64,
```
haraldng, line 77: Why is this needed? Isn't it sync_idx + suffix.len() + 0 or 1 (if StopSign is some)? I.e., we can calculate this by calling `self.internal_storage.get_accepted_idx()` at the receiver.

It is probably equally correct to explicitly send the accepted_idx. But I prefer if we use the `self.internal_storage.get_accepted_idx()` for consistency (and avoid any potential bugs)

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -119,12 +119,19 @@ where
                             "storage error while trying to read old log or snapshot",
                         );
                     }
+                    // Compact index of snapshot depends on if incoming stopsign is decided
+                    let new_compact_idx = match accsync.stopsign {
+                        Some(_) if accsync.decided_idx == accsync.accepted_idx => {
```
haraldng, line 124:
```suggestion
                        Some(ss) if accsync.decided_idx == ss.log_idx => {
```

omnipaxos/src/messages.rs
```diff
@@ -73,6 +73,8 @@ pub mod sequence_paxos {
         pub sync_idx: u64,
```
haraldng, line 73: To avoid confusion I think we should replace the `decided_snapshot, suffix, sync_idx` fields with the following:
```
enum SyncItem {
        Entries(u64, Vec<T>), // sync_idx + suffix (receiver applies suffix at sync_idx)
        SnapshotAndEntries(SnapshotType<T>, Vec<T>) // receiver applies snapshot (with decided_idx) and appends suffix)
}

```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -167,26 +169,19 @@ where
...
-                        }
-                    } else {
-                        self.accept_stopsign(ss);
-                    }
-                    accepted.accepted_idx = self.internal_storage.get_accepted_idx();
-                }
-                None => self.forward_pending_proposals(),
+            if accsync.stopsign.is_none() {
+                self.forward_pending_proposals();
+                //TODO: foward pending stopsign like in handle_majority_promises?
```
haraldng, line 176: no need to forward StopSign here. In handle_majority_promises we need to adopt the StopSign because it might be chosen. Here it is not chosen so don't need to do it.

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -214,23 +214,22 @@ where
...
+                        for follower in self.leader_state.get_promised_followers() {
+                            let decided_idx = self.internal_storage.get_decided_idx();
+                            if self.leader_state.get_accepted_idx(follower)
+                                != self.internal_storage.get_accepted_idx()
+                            {
+                                self.send_accept_stopsign(follower, ss.clone(), true);
+                            }
+                            if self.internal_storage.stopsign_is_decided() {
+                                self.send_decide(follower, decided_idx, true);
                             }
```
haraldng, line 229: are the send_accept_stopsign and send_decide here in the wrong order (i.e., the sends are inside the wrong if)?

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -214,23 +214,22 @@ where
...
-                        if !ss.decided(decided_idx) {
-                            for follower in self.leader_state.get_promised_followers() {
-                                // resend if the follower has not accepted the stopsign
-                                if !ss.decided(self.leader_state.get_accepted_idx(follower)) {
-                                    self.send_accept_stopsign(follower, ss.stopsign.clone(), true);
-                                }
+                        for follower in self.leader_state.get_promised_followers() {
+                            let decided_idx = self.internal_storage.get_decided_idx();
+                            if self.leader_state.get_accepted_idx(follower)
+                                != self.internal_storage.get_accepted_idx()
```
haraldng, line 223: Just to understand what we're doing here. Say if the leader has accepted_idx = 10 and decided_idx = 8, and follower has accepted_idx = 9. Is the idea that by resending `Decide<9>` that the follower will detect a sequence break and then catch up?

haraldng (changes requested):
Please check comments and questions. Also, if I understand correctly, we remove the log index from the StopSign and instead calculate it instead?

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -217,16 +217,16 @@ where
...
                 if *phase == Phase::Accept {
                     if let Some(ss) = self.internal_storage.get_stopsign() {
+                        let decided_idx = self.internal_storage.get_decided_idx();
                         for follower in self.leader_state.get_promised_followers() {
-                            let decided_idx = self.internal_storage.get_decided_idx();
-                            if self.leader_state.get_accepted_idx(follower)
+                            if self.internal_storage.stopsign_is_decided() {
+                                self.send_decide(follower, decided_idx, true);
+                            }
+                            else if self.leader_state.get_accepted_idx(follower)
```
kevin-harrison, line 225: If we are resending a Decide for a StopSign then our sequence number has progressed past when we sent the AcceptStopSign. Since we don't cache the AcceptStopSign's sequence number if we resent it we would just trigger a resync.

Now if a follower is missing the accept and decide we still trigger a resync when we resend the decide but send less messages (and dont send an AcceptStopSign with a false sequence number).

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -167,26 +162,18 @@ where
...
-                    }
-                    accepted.accepted_idx = self.internal_storage.get_accepted_idx();
-                }
-                None => self.forward_pending_proposals(),
+            if accsync.stopsign.is_none() {
+                self.forward_pending_proposals();
             }
+            self.internal_storage
+                .set_stopsign(accsync.stopsign)
+                .expect("storage error while trying to write stopsign");
```
haraldng, line 172: somewhere in this function, we need to set the decided_idx set by the leader.

haraldng (changes requested):

omnipaxos/src/sequence_paxos/follower.rs
kevin-harrison: This happens on line 102

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -219,43 +219,52 @@ where
...
                         .expect("storage error while trying to read diff snapshot");
                     let suffix = self
                         .internal_storage
                         .get_suffix(my_decided_idx)
                         .expect("storage error while trying to read log suffix");
-                    (Some(delta_snapshot), suffix, follower_decided_idx)
+                    let compacted_idx = match self.internal_storage.stopsign_is_decided() {
+                        true => my_decided_idx - 1,
+                        false => my_decided_idx,
+                    };
```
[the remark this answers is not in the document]
kevin-harrison: The problem is we are creating a snapshot diff for the follower and the leader doesn't adopt this diff themself. Because the leader doesn't adopt the snapshot we can't use the leader's `get_compacted_idx()`,  and instead calculate it (snapshot always goes up to leaders decided (-1 if decided stopsign)).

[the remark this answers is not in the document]
kevin-harrison: If the leader would adopt the snapshot here it would have to create the snashot diff (follower_dec, leader_dec) for the follower, then create and adopt a snapshot (leader_compactIdx, leader_dec), and then we could rely on the leader's `get_compacted_idx()`.

omnipaxos/src/messages.rs
kevin-harrison: I think of the AcceptSync as 3 separate steps (adopt Snapshot, adopt suffix, then adopt StopSign). Trying to keep these updates logically separated will make it easier to define the atomic transaction in the future.

omnipaxos/src/storage.rs
```diff
@@ -693,19 +693,22 @@ where
         }
     }
 
+    fn create_decided_snapshot(&mut self) -> StorageResult<T::Snapshot> {
+        self.create_snapshot(self.get_decided_idx())
```
kevin-harrison, line 697: This is incorrect because we don't take into account if a stopsign is included in the decided entries. It so we need to snapshot 1 less item.

omnipaxos/src/sequence_paxos/leader.rs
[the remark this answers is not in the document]
haraldng: Right, that's true.

omnipaxos/src/messages.rs
haraldng: Yes, but I think this SyncItem would help with that. Because then when matching the different variants we would potentially rollback what is required. 

I don't like that we send a compacted index (that is 0) in the cases we don't even send a snapshot, and similarly a sync index when we use snapshots.

omnipaxos/src/storage.rs
```diff
@@ -704,14 +712,25 @@ where
...
+            true => self.get_decided_idx() - 1,
+            false => self.get_decided_idx(),
+        };
+        let compacted_idx = self.get_compacted_idx();
+        if from_idx <= compacted_idx {
+            if compacted_idx < log_decided_idx {
+                Ok(Some(SnapshotType::Complete(
+                    self.create_snapshot(log_decided_idx)?,
+                )))
+            } else {
```
kevin-harrison, line 726: This is an optimization because when `compacted_idx == log_decided_idx` our current snapshot is up to date and we can just use it. 

However, it also necessary for when an "empty" snapshot doesn't  act like an identity element, like with `LatestValue` in the tests. This is because if we call `create_snapshot` when our log consists of entirely a snapshot, we will it will merge an empty snapshot onto the current one.

omnipaxos/src/storage.rs
```diff
@@ -693,23 +697,48 @@ where
...
+
     pub(crate) fn get_snapshot(&self) -> StorageResult<Option<T::Snapshot>> {
         self.storage.get_snapshot()
     }
 
+    // Creates a Delta snapshot of entries from `from_idx` to the end of the decided log and also
+    // returns the compacted idx of the created snapshot. If the range of entries contains entries
+    // which have already been compacted a valid delta cannot be created, so creates a Complete
+    // snapshot of the entire decided log instead.
     pub(crate) fn create_diff_snapshot(
```
haraldng, line 716: I'm a bit confused regarding this function. It's called create *diff* but we seem to return `Complete` snapshots?

omnipaxos/src/storage.rs
```diff
@@ -682,6 +682,10 @@ where
 
     pub(crate) fn create_snapshot(&mut self, compact_idx: u64) -> StorageResult<T::Snapshot> {
         let compacted_idx = self.get_compacted_idx();
+        assert!(
+            compact_idx >= compacted_idx,
+            "Can't create snapshot, index {compact_idx} already trimmed"
+        );
```
haraldng, line 688: I prefer if we return a StorageResult Error instead of having this assert.

haraldng (changes requested):

omnipaxos/src/storage.rs
kevin-harrison: Imagine the follower has decided_idx = 2 and the leader decided_idx = 10 but also compacted_idx 5. The follower needs a snapshot representing entries 2-10, but the leader can't create that diff since it has compacted entries 1-5, so the best it can do is merge its own snapshot(1-5) with snap(6-10) = snap(1-10). Its a Complete snapshot because the follower needs to adopt snap(1-10) as is and not merge it onto its own snapshot like it would with a delta.

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -205,57 +205,44 @@ where
...
+                    .create_diff_snapshot(followers_decided_idx)
                     .expect("storage error while trying to read diff snapshot");
                 let suffix = self
                     .internal_storage
                     .get_suffix(my_decided_idx)
                     .expect("storage error while trying to read log suffix");
-                (Some(delta_snapshot), suffix, follower_decided_idx)
+                (delta_snapshot, suffix, compacted_idx)
+            } else if followers_promise_n == max_promise_n {
+                // Follower's accepted entries are valid, send any new entries after
```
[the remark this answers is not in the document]
kevin-harrison: Doesn't checking the if with decided_idxs + snapshot first solve this?

There is also the case of followers doing crash recovery where they PrepareRequest the same leader after having crashed for a long time (but they would have `followers_promise_n == my_ballot`), maybe worth adding?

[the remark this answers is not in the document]
kevin-harrison: Ok now I think i understand what you mean, we can avoid the trade offs of having the snapshot check first vs the ballots check first by checking for ballots first and then inside that if checking for snapshots again (similarly to how it was before).

omnipaxos/src/storage.rs
haraldng: Right, makes sense

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -217,9 +217,11 @@ where
...
             .get_decided_idx(*pid)
             .expect("Received PromiseMetaData but not found in ld");
-        let (delta_snapshot, suffix, sync_idx) =
-            if followers_decided_idx < my_decided_idx && T::Snapshot::use_snapshots() {
-                // Send missing decided entries in snapshot and accepted entries in suffix
+        let (delta_snapshot, suffix, sync_idx) = if followers_promise_n == max_promise_n {
+            // Follower's accepted entries are valid, send any new entries after
+            if T::Snapshot::use_snapshots() && *followers_accepted_idx < my_decided_idx {
+                // Note: we snapshot from follower's decided and not follower's accepted because
+                // snapshots currently can't handle merging onto accepted entries.
```
kevin-harrison, line 224: We don't do the most optimal thing here which would be to send the follower a snapshot of entries (follower_acc_idx, leader_dec_idx) since the followers accepted entries are actually decided the follower just doesn't know it yet. However we would have to send some flag telling the follower to merge onto its accepted entries instead of onto its decided entries (`SnapshotType::Delta(Snapshot, bool)` or something). It makes the snapshot code messier than it already is.

Snapshotting like this is only meaningfully better than the current snapshotting when the follower has received a bunch of accepts from the previous leader but not their corresponding decides. With how sequence numbers work this shouldn't really happen.

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -205,58 +205,59 @@ where
...
-                        .internal_storage
-                        .get_suffix(*promise_accepted_idx)
-                        .expect("storage error while trying to read log suffix");
-                    (None, sfx, *promise_accepted_idx)
-                }
-            } else if follower_decided_idx < my_decided_idx && T::Snapshot::use_snapshots() {
-                let delta_snapshot = self
+        let (delta_snapshot, suffix, sync_idx) = if (followers_promise_n == max_promise_n) || (*followers_promise_n == self.leader_state.n_leader) {
+            // Follower's accepted entries are valid, send any new entries after
+            if T::Snapshot::use_snapshots() && *followers_accepted_idx < my_decided_idx {
```
haraldng, line 222: Is the second check really needed? Because we anyway send the diff snapshots from the follower's decided index.

```suggestion
            if T::Snapshot::use_snapshots() {
```

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -205,58 +205,59 @@ where
...
                 let suffix = self
                     .internal_storage
                     .get_suffix(my_decided_idx)
                     .expect("storage error while trying to read log suffix");
-                (Some(delta_snapshot), suffix, follower_decided_idx)
+                (delta_snapshot, suffix, compacted_idx)
             } else {
+                let sfx = self
+                    .internal_storage
+                    .get_suffix(*followers_accepted_idx)
```
haraldng, line 237: This is incorrect. With the check of just `if (followers_promise_n == max_promise_n) || (*followers_promise_n == self.leader_state.n_leader)` before here, it becomes incorrect to send from the followers accepted index. Consider the following case: the old leader, call it server A, disconnects with the other servers at ballot `n=1` and with 10 log entries. The other servers elect a new leader at `n=2` and decides 10 new entries (i.e., entry 0-9 were from `n=1`, and 10-19 were from `n=2`). But during the partition, A continued in `n=1` and appended 5 **other** entries (it will append because it thinks it's the leader). Now when it reconnects in `n=2`, it will keep those 5 entries that were added during the partition, which is incorrect (because its log entries from 10-14 will be from `n=1` instead of `n=2`).

Instead, we should send the `max_accepted_idx` (which you removed). It is only up to that point in the log that we know that this follower's log is valid.
```suggestion
                    .get_suffix(max_accepted_idx)
```

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -205,58 +205,59 @@ where
...
+                    .expect("storage error while trying to read log suffix");
+                (None, sfx, *followers_accepted_idx)
+            }
+        } else {
+            // Follower's accepted entries are not valid, send everying after follower's decided
```
haraldng, line 242:
```suggestion
            // Follower's accepted entries are not valid. Synchronize by sending a snapshot from the follower's decided index up to leader's decided index and any suffix.
```

haraldng (changes requested):

omnipaxos/src/sequence_paxos/leader.rs
kevin-harrison: It is not needed for safety but is an optimization (but we need to use `max_accepted_idx` not `followers_accepted_idx` as you've pointed out). The idea is that if the follower has valid accepted entries >= to the leader's `decided_idx` then we don't have to send them a snapshot of the decided entries. We just need to move the follower's `decided_idx` and potentially send a suffix of new accepted entries, and this is exactly what will happen in the `else` statement.

omnipaxos/src/sequence_paxos/leader.rs
kevin-harrison: Yeah you are right. If the follower was synced in the previous round we can only trust up to the `max_accepted_idx` of last round. And if the follower was synced by us then we can trust up to their `accepted_idx`.

omnipaxos/src/sequence_paxos/leader.rs
haraldng: Right, makes sense.

omnipaxos/src/storage.rs
```diff
@@ -629,6 +629,13 @@ where
         self.state_cache.decided_idx
     }
 
+    fn get_log_decided_idx(&self) -> u64 {
```
haraldng, line 632:
```suggestion
    fn get_decided_idx_without_stopsign(&self) -> u64 {
```

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -204,11 +204,14 @@ where
 
     fn send_accsync(&mut self, to: NodeId) {
         let my_decided_idx = self.get_decided_idx();
+        let my_n = self.leader_state.n_leader;
```
haraldng, line 207:
```suggestion
        let current_n = self.leader_state.n_leader;
```

haraldng (changes requested):
Looks mostly good. Just have a few suggestions for naming. I'll do a detailed pass after renaming.

haraldng:
@kevin-harrison since we are changing when we consider a StopSign decided, we might as well fix a related issue which is that currently, we don't know if a StopSign is decided when we perform `read()`. Perhaps we can add a `decided: bool` flag in the `LogEntry::StopSign` variant [here](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/util.rs#L246).

kevin-harrison:
> @kevin-harrison since we are changing when we consider a StopSign decided, we might as well fix a related issue which is that currently, we don't know if a StopSign is decided when we perform `read()`. Perhaps we can add a `decided: bool` flag in the `LogEntry::StopSign` variant [here](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/util.rs#L246).

Do you prefer changing the enum to `StopSign(StopSign, u64)` and the user has to check with the decided index to determine if its decided. Or do you prefer doing it for the user by the enum into two cases `DecidedStopSign(StopSign)` and `UndecidedStopSign(StopSign)`

docs/omnipaxos/reconfiguration.md
```diff
@@ -21,7 +21,7 @@ let decided_entries: Option<Vec<LogEntry<KeyValue>>> = omnipaxos.  read_decided_
 if let Some(de) = decided_entries {
     for d in de {
         match d {
-            LogEntry::StopSign(stopsign) => {
+            LogEntry::StopSign(stopsign, _stopsign_is_decided) => {
```
haraldng, line 24: This is correct since we are doing read_decided_suffix before. But for the sake of being clear in the docs, we should emphasize that a new instance should only be started when the stopsign is really decided. So here we should do a check that this flag is true first.

omnipaxos/src/util.rs
```diff
@@ -245,8 +245,9 @@ where
     Trimmed(TrimmedIndex),
     /// The entry has been snapshotted.
     Snapshotted(SnapshottedEntry<T>),
-    /// This Sequence Paxos instance has been stopped for reconfiguration.
-    StopSign(StopSign),
+    /// This Sequence Paxos instance has been stopped for reconfiguration. The accompanying bool
+    /// indicates whether the reconfiguration has been decided or not.
```
haraldng, line 249:
```suggestion
    /// This Sequence Paxos instance has been stopped for reconfiguration. The accompanying bool
    /// indicates whether the reconfiguration has been decided or not. If it is `true`, then the OmniPaxos instance for the new configuration can be started.
```

haraldng (changes requested):
Fix minor things related to comments and docs example. I will now do a pass of all changes in the whole PR.

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -214,23 +214,21 @@ where
     pub(crate) fn resend_message_timeout(&mut self) {
         match &self.state {
             (Role::Leader, phase) => {
```
haraldng, line 216: Do we also need to resend `Prepare` if we are in the prepare phase? I'm thinking in the case where we elect a new leader that sends a prepare to everybody which gets dropped. All servers are still connected so they won't elect a new leader, but we will not make progress because the leader does not complete the prepare phase.

If so then change this code to:
```rust
match &self.state {
     (Role::Leader, Phase::Prepare) => { // resend prepare }
     (Role::Leader, Phase::Accept) => { // your current code }

}
```

omnipaxos/tests/atomic_storage_test.rs
```diff
@@ -526,60 +525,3 @@ fn atomic_storage_majority_promises_test() {
         run_single_test(i);
     }
 }
-
-#[test]
-#[serial]
-fn atomic_storage_majority_accepted_stopsign_test() {
```
haraldng, line 532: These tests are not needed anymore?

omnipaxos/tests/sync_test.rs
haraldng, line 1: Please add small description of each test.

haraldng (commented):
Looks good. @kevin-harrison please check comments and modify if needed.

omnipaxos/src/sequence_paxos/mod.rs
kevin-harrison: The code now has the leader resending `Prepare`s  no matter the phase, so long as the `Role` is `Leader`. So this case will be covered. Another question is whether we want the leader to send prepares when their `Phase` is `Recovery` or `None`.

kevin-harrison: Maybe its more clear to explicitly pattern match on the entire state like you suggest?

omnipaxos/tests/atomic_storage_test.rs
kevin-harrison: The purpose of the test was to check for the atomicity of updating `decided_idx` and the result of the `ss.decided(idx)` function. It was a strange test because if you need to roll back the `decided_idx` you dont have to rollback anything about the stopsign (even when it included the `log_idx` field). But I mainly removed it because the `ss.decided(idx)` function no longer exists.

omnipaxos/src/sequence_paxos/mod.rs
haraldng: Right, now I remember. Let's keep it this way, because even in the Accept phase, we might need to send `Prepare` to the followers that haven't promised yet.

haraldng (approved):

PR #111, opened 2023-06-29
Integration test

TimXLHan:

## Issues

This PR added an integration test in the tests of OmniPaxos library.

## Breaking Changes

- Added integration_test.rs in /tests
- Changed test common in check.sh
- Updated the document to fix the error of the code examples


omnipaxos/tests/config/storage/node1/00000000000000000000.index
haraldng, line 1: Please remove all these files from this PR

omnipaxos/tests/integration_test.rs
```diff
@@ -0,0 +1,274 @@
...
+    };
+
+    let storage = MemoryStorage::default();
+    let mut omni_paxos: OmniPaxos<KeyValue, MemoryStorage<KeyValue>> =
+        omnipaxos_config.build(storage).unwrap();
+
+    // Original: https://omnipaxos.com/docs/omnipaxos/omnipaxos/
+    // Use case of creating a node with TOML config
+    let config_file_path = "tests/config/node1.toml";
+    let omnipaxos_config = OmniPaxosConfig::with_toml(config_file_path).unwrap();
```
haraldng, line 90: I think we should have each code example from the website as their own function with a link to the corresponding docs file and section.

In this way, if a function is changed or does not compile, we know exactly where in the docs to update. 

So here it could be like this
```
// This corresponds to the example:
// https://github.com/haraldng/omnipaxos/blob/master/docs/omnipaxos/index.md#creating-a-node
fn creating_a_node() {
     ...
}
```

haraldng (changes requested):
Please update the structure of the code so that we use one function per code example in the docs. And each function should have a comment with a link to the exact file and section that it corresponds to (see example in my comment)

Also remove the binary files that got pushed.

TimXLHan:
Finished code structure change.

Issue #103, opened 2023-06-21
Dashboard


TimXLHan:
Add a dashboard as a feature for omnipaxos like:
![image](https://github.com/haraldng/omnipaxos/assets/113171246/8a73bb12-4cdd-47b3-bf33-e2b2a4ade242)

PR #114, opened 2023-08-07
UI

TimXLHan:

## Issues

This PR Fix https://github.com/haraldng/omnipaxos/issues/103

## Breaking Changes

- Added UI module with feature flag
- Added public function in `omnipaxos.rs` for user to show the dashboard
- Added configuration in `omnipaxos_config` for UI

## Other Changes

- Moved the feature flags of `Looger` and `UI` to the `mod.rs`
- Changed the `ballots` in the BLE to `ballots_temp` and `ballots` to hold the ballots in different period.
- Added some get functions in BLE to get the internal states of omnipaxos for UI to dispaly


examples/ui/Cargo.toml
haraldng, line 1: Do we need a separate ui example crate? It is identical to the kv example. Is it possible to simply run the kv example with the `ui` feature instead?

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -74,7 +74,9 @@ pub(crate) struct BallotLeaderElection {
     peers: Vec<NodeId>,
     /// The current round of the heartbeat cycle.
     hb_round: u32,
-    /// Vector which holds all the received heartbeats
+    /// Vector which temporarily holds all the received heartbeats from one heartbeat round, including the current node.
+    ballots_temp: Vec<(Ballot, Connectivity)>,
+    /// Vector that holds all the received heartbeats from the previous heartbeat round, including the current node. Represents nodes that are currently alive from the view of the current node.
     ballots: Vec<(Ballot, Connectivity)>,
```
haraldng, line 80: I would say we should keep the same field and name for variables as before, and anything we need for the UI we add under the `ui` feature flag with a clear name. So maybe here you need something like
```suggestion
    /// Vector which temporarily holds all the received heartbeats from one heartbeat round, including the current node.
    #[cfg(feature = "ui")]
    prev_round_ballots: Vec<(Ballot, Connectivity)>,
    /// Vector that holds all the received heartbeats from the previous heartbeat round, including the current node. Represents nodes that are currently alive from the view of the current node.
    ballots: Vec<(Ballot, Connectivity)>,
```

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -225,15 +229,17 @@ impl BallotLeaderElection {
     }
 
     pub(crate) fn hb_timeout(&mut self) -> Option<Ballot> {
-        let my_connectivity = self.ballots.len() + 1;
+        let my_connectivity = self.ballots_temp.len() + 1;
         self.connectivity = my_connectivity as Connectivity;
+        self.ballots_temp
+            .push((self.current_ballot, self.connectivity));
+        self.ballots = std::mem::take(&mut self.ballots_temp);
```
haraldng, line 236: Similarly here: I don't think we should change the core BLE code. Instead, add what you need under the `ui` feature

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -275,6 +280,15 @@ impl BallotLeaderElection {
...
     }
+
+    pub(crate) fn get_current_ballot(&self) -> Ballot {
+        self.current_ballot
+    }
+
+    pub(crate) fn get_ballots(&self) -> &Vec<(Ballot, Connectivity)> {
+        &self.ballots
+    }
+
```
haraldng, line 291: Add ui feature flag for all functions that are only used by the UI code.

omnipaxos/src/utils/mod.rs
```diff
@@ -1,2 +1,6 @@
 /// Holds helpful functions used in creating loggers.
+#[cfg(feature = "logging")]
 pub mod logger;
+/// Holds code for ui implementation.
+#[cfg(feature = "ui")]
+pub mod ui;
```
haraldng, line 6: I would like `omnipaxos` to only hold the core logic/library, so I suggest moving this ui into its own workspace instead e.g., `omnipaxos_ui`. It would be more clean and you can simply use it as a dependency for the UI feature.

haraldng (changes requested):
Looks mostly good, but some changes are needed:

1. Modify/revert BLE code so that the UI doesn't change the original code and only adds what it needs under the `ui` feature flag.
2. Create `omnipaxos_ui` workspace instead of having it in `omnipaxos/src/utils/ui`
3. Remove the UI example and add a feature flag in the kv example instead if possible.

examples/kv_store/Cargo.toml
```diff
@@ -2,14 +2,19 @@
 name = "kv_store"
 version = "0.1.0"
 edition = "2021"
-
+default-run = "kv_store"
```
haraldng, line 5:
```suggestion
```

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -225,15 +233,17 @@ impl BallotLeaderElection {
     }
 
     pub(crate) fn hb_timeout(&mut self) -> Option<Ballot> {
-        let my_connectivity = self.ballots.len() + 1;
+        // Add our own ballot to the list of received ballots of current hb round
+        self.ballots.push((self.current_ballot, self.connectivity));
+        self.prev_round_ballots = self.ballots.clone();
+        let my_connectivity = self.ballots.len();
```
haraldng, line 239: need to update own connectivity **before** pushing it into ballots.

omnipaxos/src/util.rs
```diff
@@ -401,3 +401,16 @@ impl Quorum {
         }
     }
 }
+
+/// The states of an OmniPaxos instance.
+pub struct OmniPaxosStates{
```
haraldng, line 406:
```suggestion
/// The states to show in the UI.
pub struct UIState {
```

omnipaxos/tests/config/test.toml
```diff
@@ -64,3 +64,9 @@ storage_type = { type = "Broken" }
 [sync_test]
 wait_timeout_ms = 5000
 num_nodes = 3
+
+[ui_test]
+wait_timeout_ms = 5000
+num_nodes = 3
+num_threads = 8
+election_timeout_m = 100
```
haraldng, line 72: We have too many variants of these test configurations. If you only need something similar to the default values, then use `TestConfig::default()` in the code instead.

https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/tests/utils.rs#L96

omnipaxos_ui/src/app.rs
```diff
@@ -0,0 +1,76 @@
...
+impl PartialEq for Node {
+    fn eq(&self, other: &Self) -> bool {
+        self.pid == other.pid
+            && self.configuration_id == other.configuration_id
+            && self.ballot_number == other.ballot_number
+    }
+}
+
+/// Ui application, containing the ui states
+pub struct App {
```
haraldng, line 22: We should hide the internal state as much as possible. I don't think the fields and `set_decided_idx` need to be public. The only things that should be visible for the users are the public functions in OmniPaxosUI that are needed to create, tick, and update the ui.

haraldng (changes requested):
This is nice! Another idea that just came up is if the dashboard can have an additional widget/show more if the node is the leader. For example some information from the [LeaderState](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/util.rs#L59) that shows which servers have promised, how much they are lagging behind in the log.

TimXLHan:
The code can be reviewed, now the dashboard looks like this:
image

But there is one bug need to be fixed in next push, will look it later:
```sh
error[E0277]: the trait bound `Logger: Deserialize<'_>` is not satisfied
    --> omnipaxos/src/omni_paxos.rs:181:24
     |
181  |     pub custom_logger: Option<slog::Logger>,
     |                        ^^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `Logger`
     |
     = help: the following other types implement trait `Deserialize<'de>`:
               bool
     |        ---------- required by a bound in this associated function
1867 |     where
1868 |         V: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `MapAccess::next_value`
For more information about this error, try `rustc --explain E0277`.
error: could not compile `omnipaxos` (lib) due to 2 previous errors

**Comment by haraldng** [comment]
The PR looks mostly good. The error is because we derive a deserializer so that we can automatically deserialize the `ServerConfig` from a file. But for the custom logger that will not be possible. You might be able to make the compiler happy by telling serde to skip that field using [this](https://serde.rs/field-attrs.html#skip_deserializing). Otherwise we might need to remove it from the ServerConfig and set the custom_logger somewhere else.

File: examples/kv_store/src/util.rs
Comment (haraldng) [inline examples/kv_store/src/util.rs:1]: call all UI related constants `UI...` e.g., `UI_TICK_PERIOD` and `UI_WAIT...` and put them under the feature flag.

File: examples/kv_store/src/util.rs [hunk 10]
```diff
@@ -2,8 +2,11 @@ use std::time::Duration;
 
 pub const BUFFER_SIZE: usize = 10000;
 pub const ELECTION_TICK_TIMEOUT: u64 = 5;
-pub const TICK_PERIOD: Duration = Duration::from_millis(10);
+pub const OMNI_PAXOS_TICK_PERIOD: Duration = Duration::from_millis(10);
```
haraldng, line 5:
```suggestion
pub const TICK_PERIOD: Duration = Duration::from_millis(10);
```

omnipaxos/Cargo.toml
```diff
@@ -20,6 +20,7 @@ slog-async = { version = "2.7.0", optional = true }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 toml = { version = "0.7.3", optional = true }
 omnipaxos_macros = { version = "0.1.2", optional = true }
+#omnipaxos_ui = { path = "../omnipaxos_ui", optional = true}
```
haraldng, line 23:
```suggestion
```

omnipaxos/Cargo.toml
```diff
@@ -38,6 +39,7 @@ logging  = ["slog", "slog-term", "slog-async"]
 toml_config = ["serde", "toml"]
 default = ["batch_accept"]
 macros = ["omnipaxos_macros"]
+ui = ["logging"]
```
haraldng, line 42: is this needed?

omnipaxos/src/util.rs
```diff
@@ -401,3 +401,51 @@ impl Quorum {
         }
     }
 }
+
+/// Structs for the UI
+pub mod ui {
```
haraldng, line 406: This file is supposed to be util functions and structs related to consensus logic. Other utils should be in the `utils` mod, so move this whole module to new file in utils/ui.rs

omnipaxos/src/omni_paxos.rs
```diff
@@ -23,6 +24,7 @@ use std::{
 };
 #[cfg(feature = "toml_config")]
 use toml;
+// use crate::valid_config;
```
haraldng, line 27: remove

omnipaxos_ui/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn update(&mut self) {
+        // Handle user input
+        if crossterm::event::poll(Duration::from_millis(0)).unwrap() {
+            if let Event::Key(key) = crossterm::event::read().unwrap() {
+                if let KeyCode::Esc = key.code {
+                    self.stop();
+                };
+                if let KeyCode::Char('q') = key.code {
+                    self.stop();
+                };
```
haraldng, line 86: In the title of the dashboard, we should write `OmniPaxos node <pid> (press 'q' or 'esc' to exit the dashboard)`

omnipaxos_ui/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+                    for (idx, &accepted_idx) in
+                        op_states.cluster_state.accepted_indexes.iter().enumerate()
+                    {
+                        self.app.followers_progress[idx] = if leader_acc_idx == 0 {
+                            0.0 // To avoid division by zero
+                        } else {
+                            accepted_idx as f64 / leader_acc_idx as f64
+                        };
+                        self.app.followers_accepted_idx[idx] = accepted_idx;
+                    }
```
haraldng, line 126: move to own function `update_progressbar()`

omnipaxos_ui/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+                .iter()
+                .filter(|(b, _)| b.pid != self.app.current_node.pid)
+                .for_each(|(b, c)| {
+                    self.app.active_peers.push((*b).into());
+                    self.app.active_peers.last_mut().unwrap().connectivity = *c;
+                });
+            // Sort the active peers by pid
+            self.app.active_peers.sort_by(|a, b| a.pid.cmp(&b.pid));
+            self.app.current_node.connectivity = self.app.active_peers.len() as u8 + 1;
+            self.update();
```
haraldng, line 146: move to own function `update_active_peers()`

docs/omnipaxos/dashboard.md
haraldng, line 1: Need to describe what the dashboard is (i.e., the view from a single node), when and why it can be used. And also more explanatory in what the tick does so that users know how often they should call it.

haraldng (changes requested):
Code looks mostly good, can be restructured a bit for clearness (see comments). 

The docs need to be more descriptive on what the dashboard is, what it shows. Right now it just says how to use it.

examples/kv_store/src/util.rs
TimXLHan: could not make it under feature flg:
```rust
pub const UI_TICK_PERIOD: Duration = Duration::from_millis(200);
```
because the `tokio::select!` dose not supply feature inside, so the ui tick need to be here no matter ui started or not:
```rust
// examples/kv_store/src/server.rs
loop {
            tokio::select! {
                biased;
               // #[cfg(feature = "with_omnipaxos_ui")]    // this is not supplied
                _ = op_ui_tick_interval.tick() => {
                    #[cfg(feature = "with_omnipaxos_ui")]
                    self.omni_paxos_ui.tick(self.omni_paxos.lock().unwrap().get_ui_states());
                },
            }
        }
```

omnipaxos/Cargo.toml
TimXLHan: deleted

TimXLHan:
need to update the doc with next commit

haraldng (changes requested):
The changes mostly look good. But the `omnipaxos_ui` does not compile because there are two different versions of `ratatui`: we're using v0.22 while your dependency tui-logger is using v0.23. You can probably reproduce it by doing the following in `omnipaxos_ui`:
```bash
cargo clean
cargo check
``` 
If you then run `cargo tree | grep tui` you'll see:

```
├── ratatui v0.22.0
└── tui-logger v0.9.6
    ├── ratatui v0.23.0
```

This can be fixed by using v0.23 in `omnipaxos_ui/Cargo.toml` but then we should also include the `Cargo.lock` file to prevent this in the future.

A final thing regarding the color is that we could perhaps in the Cluster information widget instead of writing `peers` we can write `nodes`. And for every node we write with a specific colored background. And when the leader changes to a certain node, we use that color e.g., node 1 is green, node 2 is blue etc.

TimXLHan:
My local repo works fine:
image
Then I guess I just include the cargo.lock

omnipaxos_ui/src/util.rs
```diff
@@ -16,7 +16,7 @@ pub(crate) mod defaults {
     pub(crate) const UI_BARCHART_GAP: u16 = 1;
     pub(crate) const UI_TABLE_CONTENT_HEIGHT: u16 = 1;
     pub(crate) const UI_TABLE_ROW_MARGIN: u16 = 1;
-    pub(crate) const COLORS: [Color; 4] = [Color::Green, Color::Blue, Color::Magenta, Color::Cyan];
+    pub(crate) const COLORS: [Color; 4] = [Color::Green, Color::Blue, Color::Cyan, Color::Magenta];
```
haraldng, line 19: Add all colors available in the library (apart from black and white) + `Color::Indexed(208)` (orange) + `Color::Indexed(211)` (pink). And try to spread them out so more contrasting colors are next to each other.

omnipaxos_ui/src/app.rs
```diff
@@ -50,31 +54,47 @@ pub(crate) struct App {
...
+            .clone()
+            .into_iter()
+            .map(|pid| Node {
+                pid,
+                color: *color_generator.next_color(),
+                ..Default::default()
+            })
+            .collect();
+        nodes.push(current_node.clone());
+        nodes.sort_by(|a, b| a.pid.cmp(&b.pid));
```
haraldng, line 84: We should first push(), then sort, and then map the colors so that everything is consistent even if we run dashboard on different nodes.

omnipaxos_ui/src/render.rs
```diff
@@ -0,0 +1,473 @@
...
+    // Table
+    let table = draw_leader_table(app, Borders::TOP | Borders::LEFT | Borders::BOTTOM);
+    f.render_widget(table, table_chunks[0]);
+    // Progress bar
+    draw_progress(f, app, table_chunks[1]);
+}
+
+fn draw_title<'a>(app: &App) -> Paragraph<'a> {
+    Paragraph::new(format!(
+        "{} Node <{:?}> (press 'q' or 'esc' to exit)",
```
haraldng, line 136:
```suggestion
        "{} node {:?} (press 'q' or 'esc' to exit the dashboard)",
```

omnipaxos_ui/src/util.rs
```diff
@@ -0,0 +1,70 @@
...
+        Self {
+            configuration_id: config.cluster_config.configuration_id,
+            pid,
+            peers,
+        }
+    }
+}
+
+pub(crate) struct ColorGenerator {
+    colors: Vec<Color>,
```
haraldng, line 52: Remove this and use the constant instead.

omnipaxos_ui/src/util.rs
```diff
@@ -0,0 +1,70 @@
...
+impl ColorGenerator {
+    pub fn new(colors: Vec<Color>) -> Self {
+        ColorGenerator { colors, index: 0 }
+    }
+
+    /// Get the next color in the list, and increase the index.
+    pub fn next_color(&mut self) -> &Color {
+        if self.colors.is_empty() {
+            panic!("No colors provided!");
+        }
```
haraldng, line 65: This can be removed when we use the constant directly.

omnipaxos_ui/src/render.rs
```diff
@@ -0,0 +1,473 @@
...
+            ]
+            .as_ref(),
+        )
+        .split(cell);
+
+    let gauge_color = if ratio == 1.0 {
+        Color::Green
+    } else {
+        Color::LightYellow
+    };
```
haraldng, line 463: 90 - 100 : Green
75 - 90: LightYellow
50 - 75: `Color::Indexed(208)` (orange)
0 - 50: LightRed

examples/kv_store/src/main.rs
haraldng, line 1: I find the main function too messy, which is not good because we want the example to be simple and clear. I suggest we go back to how the old kv_store looked like and instead create a new example crate `examples/dashboard` and move all the ui code there. 

You will probably want to reuse some setup code from the `kv_store`, so need to make it both a bin and lib: https://stackoverflow.com/questions/26946646/package-with-both-a-library-and-a-binary

In the main() of `examples/dashboard` we should make one loop/batch of entries for a few seconds (instead of just proposing one value), then kill the leader and loop again.

haraldng (changes requested):
Restructure code into separate example and add more colors.

omnipaxos_ui/src/util.rs
TimXLHan: The constant colors list may not be long enough for the number of servers. The idea of `ColorGenerator` is to make sure we always can get color form the colors list -- when reach the end, the index will point back to the first one. So I'd like to keep it.

examples/kv_store/src/util.rs
```diff
@@ -10,3 +10,4 @@ pub const WAIT_DECIDED_TIMEOUT: Duration = Duration::from_millis(50);
 #[cfg(feature = "with_omnipaxos_ui")]
 pub const UI_UPDATE_TIMEOUT: Duration = Duration::from_secs(1);
 pub const UI_TICK_PERIOD: Duration = Duration::from_millis(200);
+
```
haraldng, line 13: please do not push these unnecessary changes in git

omnipaxos_ui/src/app.rs
```diff
@@ -73,10 +72,11 @@ impl App {
         peers.sort();
         let nodes: Vec<Node> = peers
             .into_iter()
-            .map(|pid| {
+            .enumerate()
```
haraldng, line 75: sort the pids first so that the colors are the same regardless of which node/dashboard we start with.

haraldng (changes requested):
Sort the `peers` vector so that the colors will be the same regardless of which node we start the dashboard at.

omnipaxos_ui/src/app.rs
TimXLHan: It sorted at the line `72` above

examples/dashboard/src/kv.rs
haraldng, line 1: For this demo, we don't really need the KeyValue so I suggest using `u64` as the entry type for OmniPaxos (and not use snapshots).

examples/dashboard/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    for pid in SERVERS {
+        let (sender, receiver) = mpsc::channel(BUFFER_SIZE);
+        sender_channels.insert(pid, sender);
+        receiver_channels.insert(pid, receiver);
+    }
+    (sender_channels, receiver_channels)
+}
+
+fn main() {
```
haraldng, line 37: add some docs (`///`) here at the top that describes what the main does.

Then also add some throughout the function so that it is easy to understand what's going on.

examples/kv_store/Cargo.toml
```diff
@@ -2,18 +2,15 @@
 name = "kv_store"
 version = "0.1.0"
 edition = "2021"
+default-run = "kv_store"
```
haraldng, line 5: is this needed?

haraldng (changes requested):

examples/kv_store/Cargo.toml
TimXLHan: its needed if we want to run the `kv_store` by simply `cargo run`. Otherwise we need to type `cargo run --bin kv_store` . I can add some comments in the code if you want, or just remove this line?

haraldng: remove this line. If you want to run kv_store you change directory to kv_store and then do `cargo_run`

haraldng (changes requested):

examples/dashboard/src/kv.rs
```diff
@@ -1,40 +1,6 @@
...
-        for (k, v) in delta.snapshotted {
-            self.snapshotted.insert(k, v);
-        }
-    }
-
-    fn use_snapshots() -> bool {
-        true
-    }
-}
+pub struct LogEntry(pub u64);
```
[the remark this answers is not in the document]
TimXLHan:
```rust
type OmniPaxosKV = OmniPaxos<u64, MemoryStorage<u64>>;
```

image

It seems not implemented, should we keep the wrapper?

haraldng (approved):

TimXLHan:
The code worked fine on my local machine. I added changed type casting to see if it can pass the `test(nightly)` now.

haraldng (approved):

PR #117, opened 2023-08-16
BLE optimizations

kevin-harrison:

## Changes
Updates leader election. 
## Other Changes
Also adds std::time::Duration parsing to TestConfig TOML.


omnipaxos/src/ballot_leader_election.rs
```diff
@@ -75,12 +75,14 @@ pub(crate) struct BallotLeaderElection {
     /// The current round of the heartbeat cycle.
     hb_round: u32,
     /// Vector which holds all the received heartbeats
-    ballots: Vec<(Ballot, Connectivity)>,
+    ballots: Vec<(Ballot, Connectivity, bool)>,
```
haraldng, line 78: For code clarity, I think it would be nicer to, instead of using bool, have an alias for `Happy`  or perhaps even better have a struct 
```
BallotResponse {
      ballot: Ballot,
      connectivity: u8 // or Connectivity
      happy: bool
}
```

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -75,12 +75,14 @@ pub(crate) struct BallotLeaderElection {
...
     /// Holds the current ballot of this instance.
     current_ballot: Ballot,
     /// The number of replicas inside the cluster that this instance is
     /// connected to (based on heartbeats received) including itself.
     connectivity: Connectivity,
+    happy: bool,
+    gossiped_leader: Ballot,
```
haraldng, line 85: I think a better name is `highest_seen_ballot` since it can be set by ourselves (and therefore not necessarily gossiped) and can get updated before it is confirmed as a leader.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -225,15 +234,32 @@ impl BallotLeaderElection {
...
                 self.logger,
                 "Received a majority of heartbeats, round: {}, {:?}", self.hb_round, self.ballots
             );
-            self.ballots.push((self.current_ballot, self.connectivity));
+            let unhappy_nodes = self.ballots.iter().filter(|&&(_, _, happy)| !happy).count();
+            if self.quorum.is_prepare_quorum(unhappy_nodes) {
+                self.current_ballot.n = self.gossiped_leader.n + 1;
+                self.gossiped_leader = self.current_ballot;
+                self.happy = true;
+            }
```
haraldng, line 262: Can we move this to check_leader()? I think it's easier to read when all the leader changing logic is in the same function.

haraldng (changes requested):

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -119,6 +121,8 @@ impl BallotLeaderElection {
                     .unwrap_or_else(|| format!("logs/paxos_{}.log", pid));
                 create_logger(s.as_str())
             },
+            have_seen_a_quorum_of_followers: false, //TODO: what if leader wasn't
```
haraldng, line 124: To make code more clear and less error-prone we should change the `leader` field to be of type: `(Option<Ballot>, bool)` where the bool is this flag.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -266,7 +300,8 @@ impl BallotLeaderElection {
 
     fn handle_reply(&mut self, rep: HeartbeatReply) {
         if rep.round == self.hb_round && rep.ballot.config_id == self.configuration_id {
-            self.ballots.push((rep.ballot, rep.connectivity));
+            if rep.ballot.pid == self.leader.unwrap_or_default().pid {}
```
haraldng, line 303: remove?

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -63,7 +63,7 @@ where
                 });
             }
         } else {
-            self.state.0 = Role::Follower;
```
haraldng, line 66: Not sure if it is stictly necessary to remove this. We had it here before to prevent the stale leader to continue sending new entries if it has been overtaken.

haraldng (changes requested):

omnipaxos/src/sequence_paxos/leader.rs
kevin-harrison: With the new election its only the new leader's seq_paxos that triggers handle_leader(). This is because we are being cautious about exactly when to tell paxos we think we are leader (we want a majority of BLEs to see the ballot first, lest the ballot gets established in paxos and lost in BLE). The idea is that the stale leader's paxos will nearly immediately go into follower mode anyways because it will receive the new leader's Prepare message.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -224,39 +179,118 @@ impl BallotLeaderElection {
...
+                    self.pid, self.leader
+                ));
+            }
+            _ => (),
+        }
+
+        // Update happiness
+        self.happy = if self.current_ballot == self.leader.unwrap_or_default() {
+            println!("BLE {}: checking happy for leader", self.pid);
+            let followers = self
```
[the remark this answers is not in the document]
kevin-harrison: This isn't just the connectivity, it's specifically the number of nodes that think I'm their leader.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -179,116 +199,86 @@ impl BallotLeaderElection {
...
+        self.happy = match self.state {
+            BleState::Follower => {
+                let leaders_reply = self
+                    .heartbeat_replies
+                    .iter()
+                    .find(|r| r.ballot.pid == self.leader.pid);
+                match leaders_reply {
+                    Some(reply) if reply.ballot >= self.leader => reply.happy,
+                    // Must have learned of leader ballot from HeartbeatRequest and HeartbeatReply is stale
+                    Some(_) => true,
```
haraldng, line 232: I understand the case you describe in the comment. But what if the leader crashes and comes back with a smaller (the default) ballot? Do we perhaps need a check for if the `reply.ballot` is less than `self.leader` and equals the default, then become unhappy?

haraldng: On a second thought, maybe the cleanest and most similar to the pseudo code is to use another variable `leaders_latest_reply` which is the max of `self.leader` and `leader_reply`. And then set happiness based on `leaders_latest_reply`

haraldng (changes requested):

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -103,15 +121,20 @@ impl BallotLeaderElection {
...
         let mut ble = BallotLeaderElection {
             configuration_id: config_id,
             pid,
             peers,
             hb_round: 0,
             heartbeat_replies: Vec::with_capacity(num_nodes),
             current_ballot: initial_ballot,
-            happy: true, //TODO: can this be false if there is an initial leader?
-            leader: Some(initial_ballot), //TODO: is this ok
+            happy: true,
```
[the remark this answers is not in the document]
kevin-harrison: I think this is similar to when we set connectivity to the max initially. It's safe to set to true (if a leader doesn't get majority somehow it will go unhappy) and allows for an initial leader to be chosen in the first round.

omnipaxos/src/ballot_leader_election.rs
kevin-harrison: Not sure I understand the second solution, but yeah it needs to be addressed. Another solution is to make ballots persistent.

haraldng (changes requested):

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -59,8 +62,18 @@ impl PartialOrd for Ballot {
     }
 }
 
-/// The connectivity of an OmniPaxos node
-type Connectivity = u8;
+/// The election state of a BLE
+pub(crate) enum ElectionStatus {
+    /// I'm the leader.
+    Leader,
+    /// I think I'm the leader, but I see there is a greater leader that I'm not following.
```
haraldng, line 69:
```suggestion
    /// I think I'm the leader, but I know about a greater leader that others are following and I'm not connected to.
```

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -114,21 +112,20 @@ impl BallotLeaderElection {
...
+            Some(_) => Ballot::with(config_id, RECOVERY_ROUND, config.priority, pid),
+            None => Ballot::with(config_id, INITIAL_ROUND, config.priority, pid),
+        };
         let initial_leader = match recovered_leader {
-            Some(b) if b == Ballot::default() => initial_ballot,
+            Some(b) if b == Ballot::default() => {
+                Ballot::with(config_id, INITIAL_ROUND, config.priority, pid)
+            }
             Some(b) => b,
-            None => initial_ballot,
```
haraldng: instead of pattern matching twice, do it once and return a tuple `let (initial_ballot, initial_leader) = match { ... }`

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -208,80 +204,78 @@ impl BallotLeaderElection {
...
         self.new_hb_round();
-        (self.leader, self.election_status)
+        (election_status, self.leader)
     }
 
     fn update_leader(&mut self) {
         self.leader = self
             .heartbeat_replies
             .iter()
             .fold(self.leader, |a, b| a.max(b.ballot));
```
haraldng, line 225: This fold looks a bit weird, can't we just do `max_by`? 
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -87,15 +87,13 @@ pub(crate) struct BallotLeaderElection {
...
     heartbeat_replies: Vec<HeartbeatReply>,
     /// The current ballot of this instance.
     current_ballot: Ballot,
-    /// The current leader of this instance
+    /// The current leader of this instance.
     leader: Ballot,
-    /// How this instance relates to its leader
-    election_status: ElectionStatus,
-    /// If this instance sees a need for a new leader
+    /// If this instance sees a need for a new leader.
```
haraldng, line 96: It would be nice to also document the implications of the value of happy. E.g., if we see any node that is happy, it means that there is a leader somewhere in the cluster that is making progress. Unhappy means that this specific node is seeking a new leader.

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -208,80 +204,78 @@ impl BallotLeaderElection {
...
+                .filter(|hb_reply| hb_reply.leader == self.current_ballot)
+                .count();
+            let have_quorum_of_followers = match seq_paxos_state {
+                (Role::Leader, Phase::Accept) => self.quorum.is_accept_quorum(followers + 1),
+                _ => self.quorum.is_prepare_quorum(followers + 1),
+            };
+            let see_larger_ballot = self
+                .heartbeat_replies
+                .iter()
+                .any(|r| r.leader > self.current_ballot);
```
haraldng, line 271: do we also need to check r.happy here?

haraldng (changes requested):
Some small style issues and add more docs. Fix comment with question I had if required.

kevin-harrison:
The [f9b54ca](https://github.com/haraldng/omnipaxos/pull/117/commits/f9b54ca8525f5c08aedda04b04ca4684db9a26ae) commit adds the NotAccepted message so that a stale leader will remove a follower that is promised to a greater ballot from its promised followers. This can reduce traffic but the stale leader will still resend prepare messages to the unpromised followers every resend-message tick.

haraldng (changes requested):
Keep track of the `NotAccepted` servers so that we don't keep sending prepare to them upon tick timeout. Otherwise this is ready.

omnipaxos/src/util.rs
```diff
@@ -146,6 +146,11 @@ where
         self.quorum.is_prepare_quorum(num_promised)
     }
 
+    pub fn remove_promise(&mut self, pid: NodeId) {
+        self.decided_indexes[Self::pid_to_idx(pid)] = None;
+        self.promises_meta[Self::pid_to_idx(pid)] = None;
+    }
```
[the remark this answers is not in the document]
haraldng: This issue is also related and can be in this PR if you want: https://github.com/haraldng/omnipaxos/issues/118

omnipaxos/src/util.rs
```diff
@@ -178,20 +181,27 @@ where
...
+        self.promises_meta
             .iter()
             .enumerate()
-            .filter(|(pid, x)| x.is_none() && *pid != Self::pid_to_idx(self.n_leader.pid))
-            .map(|(idx, _)| (idx + 1) as NodeId)
+            .filter_map(|(idx, x)| match x {
+                Some(prom) if prom.n >= self.n_leader => None,
+                _ => Some((idx + 1) as NodeId),
+            })
+            .filter(|pid| *pid != self.n_leader.pid)
```
haraldng, line 204: same as above.

omnipaxos/src/util.rs
```diff
@@ -16,6 +16,7 @@ pub(crate) struct AcceptedMetaData<T: Entry> {
 #[derive(Debug, Clone, Default)]
 /// Promise without the suffix
 pub(crate) struct PromiseMetaData {
+    pub n: Ballot,
```
haraldng, line 19: We didn't store the promise here because the leaderstate is supposed to be cleaned when becoming leader in a new round, and all the `PromiseMetaData` are specific to this round.

Since we right now don't really do anything with nodes with a higher promise (other than not sending prepare/accept to them), I prefer if we instead keep track of them in [`promises_meta` field](https://github.com/haraldng/omnipaxos/blob/ff063b90228d6e1767c9712538bfa4c91bdc2c1e/omnipaxos/src/util.rs#L65) in `LeaderState` where `promises_meta` field is a `Vec<PromiseState>`:

```rust
enum PromiseState {
        NotPromised,
        Promised(PromiseMeta),
        PromisedHigher
}
```

I think that will make the code cleaner and less prone to introducing bugs, we don't store a bunch of ballots we don't use, and we can extend it in the future if we want to do something with those `PromisedHigher`

haraldng (changes requested):
Keep track of the nodes that have promised a higher ballot using a new enum instead of storing the promised ballot in `PromiseMetaData` (see [comment](https://github.com/haraldng/omnipaxos/pull/117#discussion_r1347371196)).

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -42,6 +42,7 @@ where
                 suffix: vec![],
                 stopsign: self.internal_storage.get_stopsign(),
             };
+            self.leader_state.reset_promises();
```
haraldng, line 45: This should not be needed since we create a new LeaderState before (on line 27).

haraldng: Can therefore also remove the reset_promises() function.

haraldng (changes requested):
Remove unnecessary reset_promises and this is good to go

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -63,14 +63,18 @@ where
...
+    pub(crate) fn become_follower(&mut self) {
+        self.state.0 = Role::Follower;
+    }
+
+    pub(crate) fn handle_preparereq(&mut self, prepreq: PrepareReq, from: NodeId) {
         #[cfg(feature = "logging")]
         debug!(self.logger, "Incoming message PrepareReq from {}", from);
-        if self.state.0 == Role::Leader {
+        if self.state.0 == Role::Leader && prepreq.n <= self.leader_state.n_leader {
             self.leader_state.set_decided_idx(from, None);
```
haraldng, line 78: We need to reset the PromiseState of this node to be `NotPromised` as well here. Otherwise we will keep sending accepts to it since the `get_promised_followers()` is now based on the `leader_state.promises_meta` rather than `leader_state.decided_indexes`.

Looking at the code now, I'm not sure why we even have the `leader_state.decided_indexes`. The decided index should probably be stored in the `PromiseMetaData`. @kevin-harrison could you please double check if there is any reason to keep the decided_indexes separate? And if not then add this change?

haraldng (changes requested):
Please fix two comments that I found during a final pass of the code.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -217,8 +218,9 @@ where
     }
 
     pub(crate) fn handle_accept_stopsign(&mut self, acc_ss: AcceptStopSign) {
-        if self.internal_storage.get_promise() == acc_ss.n
-            && self.state == (Role::Follower, Phase::Accept)
+        if self.internal_storage.get_promise() != acc_ss.n {
```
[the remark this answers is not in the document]
haraldng: then in the next else if we need to check for `==` also

omnipaxos/src/sequence_paxos/leader.rs
kevin-harrison: We take the max of `leader_state.decided_indexes` in `handle_majority_promises()` to sync with the up-to-date decided index after attaining leadership. If its possible for a node to have the greatest decided_idx but not the greatest `PromiseMetaData` then it makes since to have these two separate.

omnipaxos/src/sequence_paxos/follower.rs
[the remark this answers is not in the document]
kevin-harrison: If `acc.n` < the promise, then we want the leader to know its ballot is out-competed. The `acc.n` > the promise scenario really shouldn't happen. A leader will only send non-prepare messages to followers who have promised to it. This means the followers promise will be >= the ballot of any message sent by a leader. In case this does somehow happen we could send a PrepareReq just to be safe?

omnipaxos/src/sequence_paxos/leader.rs
haraldng: right, but then let's anyway make the change and store the decided index in `PromiseMetaData`. We just need to make sure that the comparator for PromiseMetaData doesn't include the decided index, and we also need a new function `get_max_decided_idx()` that takes the max decided index among PromiseMetaData.

omnipaxos/src/sequence_paxos/follower.rs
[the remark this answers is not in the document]
haraldng: You're right, larger than should not happen. And if it is equal, then it will be ignored if we're in recovery mode (in which we already sent the PrepareReq). Then perhaps we can change the if to `acc. n` < the promise since that is the only possible case.

haraldng (approved):

omnipaxos/src/ballot_leader_election.rs
```diff
@@ -79,7 +79,10 @@ pub(crate) struct BallotLeaderElection {
     hb_round: u32,
     /// The heartbeat replies this instance received during the current round.
     heartbeat_replies: Vec<HeartbeatReply>,
-    /// The current ballot of this instance.
+    /// Vector that holds all the received heartbeats from the previous heartbeat round, including the current node.
```
haraldng, line 82:
```suggestion
    /// Vector that holds all the received heartbeats from the previous heartbeat round, including the current node. Only used to display the connectivity of this node in the UI.
```

haraldng (approved):

PR #121, opened 2023-10-29
Atomic storage API

### LINKED ISSUE #49: Implement better trim functionality
The current trim() of PersistentStorage is implemented as a temporary solution. The only function that can remove elements from commitlog, `truncate(offset)` does not remove the first element as it only deletes entries after `offset`. This is still an open [issue](https://github.com/zowens/commitlog/issues/29) at Commitlog

Trim() works by getting the entries that should remain after trimming and then removing the commitlog through `std::fs`, It then recreates the commitlog and appends the entries. This operation is costly in performance and there could be better ways to trim the commitlog.

Mentioned by @haraldng in https://github.com/haraldng/omnipaxos/pull/48#discussion_r929044356_

### LINKED ISSUE #97: Proposal: support truly atomic storage implementations

kevin-harrison:

## Issues
Fix #97, Fix #49 

## Breaking Changes
 - The`Storage` trait has been reworked to facilitate atomic writes and allow for arbitrary crash-recovery. The indices passed to these functions now refer to the indices of the entry log as if no trimming or snapshotting has happened.
 - `PersistentStorage` and its config structs have been reworked to use RocksDB and implement the new `Storage` trait. The dependence on commitlog is therefore gone, addressing issue #49.
 - All indices are now of type `usize` instead of `u64`. This includes the indices used in the `Storage` trait, `Message` structs, and `Omnipaxos` public functions.
 - Both `Storage::get_promise` implementations in the `omnipaxos_storage` module now return `None` instead of the default `Ballot` when there is no promise ballot saved in storage.
 - The `batch_accept` feature flag has been removed and is now the default behavior.
 - The `Promise`, `AcceptSync`, and `AcceptDecide` message struct fields have been changed.

## Other Changes
 - `MemoryStorage` now implements the new `Storage` trait but its implementation of `write_atomically` does not guarantee atomicity (acceptable since memory can't persist state through crash-recovery).
 - Batched entries are now correctly flushed when reconfiguring and when deciding entries.
 - The syncing behavior in the the prepare phase in now consistent between both leaders and followers.


haraldng (changes requested):
Just some initial comments. I'll do a more detailed review after these comments are fixed.

For next time, please don't do refactorings before the actual feature is approved. E.g., here the changes for atomic storage wouldn't have been bloated with the changes for replacing u64 with usize (which is trivial).

haraldng:
After our discussion yesterday, I looked at the current code to see how we can keep as much as possible of the current batching and at the same time introduce transaction for atomic storage. I think what we need is a struct `InternalStorageTxn` for `InternalStorage`. This is to enable the clean transaction object API in the Sequence Paxos code. When we handle the `InternalStorageTxn` in `InternalStorage`, we will know which entries to actually flush. And based on that we will create the real transaction for storage. So something like this:

```rust
// transaction object for InternalStorage
struct InternalStorageTxn {
    state_ops: Vec<StateOp>, // StateOp is an enum for modifying accepted_round, decided_idx, etc.
    append_op: AppendOp,       // AppendOp is an enum for append, append_on_prefix, etc.
    snapshot: Option<(usize, S)>  // snapshot index and snapshot
}

// handle_acceptsync() in SequencePaxos
let state_ops = vec![StorageOp::AcceptedRound(4), StorageOp::DecidedIdx(10)];
let append_op = AppendOp::AppendOnPrefix(accsync.sync_idx, accsync.suffix);
let internal_txn = InternalStorageTxn {state_ops, append_ops, ..};
self.internal_storage.do_txn(internal_txn);

// do_txn(it: InternalStorageTxn) in InternalStorage
let append_res = match it.append_op {
    Append(entries, do_batching) if do_batching => self.state_cache.append_entries(entries),
    Append(entries, do_batching) if !do_batching => self.state_cache.append_entries_without_batching(entries),
    // ... AppendOnPrefix etc.
};
let mut txn = StorageTxn::new();  // transaction object for Storage
txn.insert(it.state_ops);
if let Some(entries_to_be_flushed) = append_res {
    txn.insert(StorageOp::Append(entries_to_be_flushed));
}
if let Some((idx, snapshot)) = it.snapshot {
    txn.insert(StorageOp::Snapshot(idx, snapshot));
}
self.state_cache.real_log_len = self.storage.flush(txn)?;
Ok(self.get_accepted_idx())
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -87,87 +89,34 @@ where
...
-                            "storage error while trying to read old log or snapshot",
-                        );
+            let txn = InternalStorageTxn::new();
+            txn.set_accepted_round(accsync.n);
+            txn.set_decided_idx(accsync.decided_idx);
+            if let Some(snap) = accsync.decided_snapshot {
+                match snap {
+                    SnapshotType::Complete(c) => txn.set_snapshot(accsync.sync_idx, c),
+                    SnapshotType::Delta(d) => {
+                        let decided_snap = self.internal_storage.create_merged_snapshot(accsync.sync_idx, d)
```
haraldng, line 101: I think this should be part of the transaction as well.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -87,87 +89,34 @@ where
...
+            txn.append_on_prefix(accsync.sync_idx, accsync.suffix);
             match accsync.stopsign {
-                Some(ss) => self.accept_stopsign(ss),
+                Some(ss) => txn.set_stopsign(Some(ss)),
                 None => self.forward_pending_proposals(),
             }
+            // Commit write and reply
+            let accepted_idx = self
+                .internal_storage
+                .execute_txn(txn)
```
haraldng, line 115: I can't see the definition of this function.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -260,38 +180,15 @@ where
...
-    ) -> StorageResult<()> {
-        let accepted_res = self
-            .internal_storage
-            .append_encoded_entries_and_get_accepted_idx(encoded_entries)?;
-        if let Some(accepted_idx) = accepted_res {
-            self.handle_flushed_accepted(n, accepted_idx);
+            if dec.decided_idx > self.internal_storage.get_accepted_idx() {
+                let new_accepted_idx = self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG);
+                self.handle_flushed_accepted(dec.n, new_accepted_idx);
+            }
```
haraldng, line 186: I think this can be embedded into the `set_decided_idx` in InternalStorage

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -250,79 +209,63 @@ where
...
-                                );
-                            }
-                        }
-                    }
-                    _ => {
-                        #[cfg(not(feature = "unicache"))]
-                        self.send_accept_and_cache(pid, am.flushed_entries.clone());
-                        #[cfg(feature = "unicache")]
-                        self.send_encoded_accept_and_cache(pid, am.flushed_processed.clone());
+            let pending_acceptdecide = match self.leader_state.get_batch_accept_meta(pid) {
```
haraldng, line 215:
```suggestion
            let cached_acceptdecide = match self.leader_state.get_batch_accept_meta(pid) {
```

omnipaxos/src/storage.rs
```diff
@@ -97,34 +97,149 @@ where
...
+}
+
 /// Trait for implementing the storage backend of Sequence Paxos.
 pub trait Storage<T>
 where
     T: Entry,
 {
+    /// Atomically perform and persist all write operations in order.
+    /// Returns an Error if the transaction was not committed (so it was rolled back).
+    fn write_batch(&mut self, batch: Vec<StorageOp<T>>) -> StorageResult<()>;
```
haraldng, line 224: rename to `do_txn` or `perform_ops_atomically` to not confuse with the batched entries.

haraldng (changes requested):

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -16,67 +16,83 @@ where
...
-                        .expect("storage error while trying to read log suffix");
-                    (delta_snapshot, suffix)
-                } else {
-                    let suffix = self
-                        .internal_storage
-                        .get_suffix(prep.accepted_idx)
-                        .expect("storage error while trying to read log suffix");
-                    (None, suffix)
-                }
+                // I'm more up to date and no split-brain possible => append onto preparer's
```
haraldng, line 58:
```suggestion
                // I'm more up to date in the same round => leader is at a prefix so send what he is missing.
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -16,67 +16,83 @@ where
...
+                            .internal_storage
+                            .get_suffix(decided_idx)
+                            .expect(READ_ERROR_MSG);
+                        (delta_snapshot, suffix, compacted_idx)
+                    } else {
+                        let suffix = self
+                            .internal_storage
+                            .get_suffix(prep.accepted_idx)
+                            .expect(READ_ERROR_MSG);
+                        (None, suffix, prep.accepted_idx)
```
haraldng, line 78: This logic is more or less the same as in the previous case but with a different index used to get the suffix. Can we create a function for it?

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -190,50 +137,31 @@ where
...
+                .expect(WRITE_ERROR_MSG);
+            if acc_dec.decided_idx > self.internal_storage.get_decided_idx() {
+                if acc_dec.decided_idx > self.internal_storage.get_accepted_idx() {
+                    new_accepted_idx =
+                        Some(self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG));
+                }
                 self.internal_storage
-                    .set_decided_idx(acc.decided_idx)
-                    .expect("storage error while trying to write decided index");
+                    .set_decided_idx(acc_dec.decided_idx)
```
haraldng, line 159: I think here we should check that the `new_accepted_idx` really was >= the decided index before setting the decided index.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -260,38 +184,17 @@ where
             && self.state.1 == Phase::Accept
             && self.handle_sequence_num(dec.seq_num, dec.n.pid) == MessageStatus::Expected
         {
+            if dec.decided_idx > self.internal_storage.get_accepted_idx() {
+                let new_accepted_idx = self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG);
+                self.handle_flushed_accepted(dec.n, new_accepted_idx);
```
haraldng, line 189: same here, check that the flushed was actually >= decided

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -260,38 +184,17 @@ where
...
-            .internal_storage
-            .append_encoded_entries_and_get_accepted_idx(encoded_entries)?;
-        if let Some(accepted_idx) = accepted_res {
-            self.handle_flushed_accepted(n, accepted_idx);
-        }
-        Ok(())
-    }
-
-    fn handle_flushed_accepted(&mut self, n: Ballot, accepted_idx: u64) {
+    fn handle_flushed_accepted(&mut self, n: Ballot, accepted_idx: usize) {
```
haraldng, line 197: To make code more intuitive when reading the message handlers, we can rename this

```suggestion
    fn reply_accepted(&mut self, n: Ballot, accepted_idx: usize) {
```

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -19,27 +22,23 @@ where
...
         if self.pid == n.pid {
             self.leader_state =
                 LeaderState::with(n, self.leader_state.max_pid, self.leader_state.quorum);
-            self.internal_storage
-                .flush_batch()
-                .expect("storage error while trying to flush batch");
+            // Flush any pending writes
+            // Don't have to handle flushed entries here because we will sync with followers
+            self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG);
+            self.internal_storage.set_promise(n).expect(WRITE_ERROR_MSG);
```
haraldng, line 31: use `let _ = ...` if we ignore any return values.

omnipaxos/src/sequence_paxos/leader.rs
haraldng: please fix

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -339,136 +293,35 @@ where
...
+                let pending_entries = std::mem::take(&mut self.pending_proposals);
+                new_accepted_idx = self
+                    .internal_storage
+                    .append_entries_without_batching(pending_entries)
+                    .expect(WRITE_ERROR_MSG);
             }
-            None => {
-                self.append_pending_proposals();
-                self.adopt_pending_stopsign();
+            if let Some(ss) = self.pending_stopsign.take() {
```
haraldng, line 314: we do a check for `self.internal_storage.get_stopsign().is_none()` in the if statement just before. This makes me wonder:
1. Do we need the `self.pending_stopsign `field at all?
2. Why don't we do the match on pending_stopsign.take() instead of doing two different ifs?

omnipaxos/src/sequence_paxos/leader.rs
```diff
@@ -514,38 +367,27 @@ where
             self.leader_state.accepted_indexes
         );
         if accepted.n == self.leader_state.n_leader && self.state == (Role::Leader, Phase::Accept) {
-            let old_decided_idx = self.internal_storage.get_decided_idx();
             self.leader_state
                 .set_accepted_idx(from, accepted.accepted_idx);
+            let old_decided_idx = self.internal_storage.get_decided_idx();
             if accepted.accepted_idx > old_decided_idx
```
haraldng, line 373:
```suggestion
            if accepted.accepted_idx > self.internal_storage.get_decided_idx()
```

omnipaxos/src/sequence_paxos/mod.rs
```diff
@@ -369,70 +299,21 @@ where
...
-        self.internal_storage
-            .set_stopsign(Some(ss))
-            .expect("storage error while trying to write stopsign");
-        if self.state.0 == Role::Leader {
-            let accepted_idx = self.internal_storage.get_accepted_idx();
-            self.leader_state.set_accepted_idx(self.pid, accepted_idx);
+        #[cfg(feature = "logging")]
+        info!(
+            self.logger,
+            "Propose reconfiguration {:?}", new_config.nodes
```
haraldng, line 308:
```suggestion
            "Accepting reconfiguration {:?}", new_config.nodes
```

omnipaxos/src/storage.rs
haraldng: please fix

omnipaxos/src/storage.rs
haraldng, line 1: I think the storage file has grown way too large. Let's move the InternalStorage and possibly the StateCache into separate files.

omnipaxos/src/storage.rs
```diff
@@ -796,12 +696,56 @@ where
...
+                }
+                None => (),
+            }
+            self.state_cache.accepted_idx = sync.sync_idx + sync.suffix.len();
+            sync_txn.push(StorageOp::AppendOnPrefix(sync.sync_idx, sync.suffix));
+            self.state_cache.stopsign = sync.stopsign.clone();
+            if sync.stopsign.is_some() {
+                self.state_cache.accepted_idx += 1;
+            }
+            sync_txn.push(StorageOp::SetStopsign(sync.stopsign));
```
haraldng, line 735: setting the stopsign must be inside the if?

omnipaxos/src/messages.rs
```diff
@@ -50,15 +50,29 @@ pub mod sequence_paxos {
...
+        pub accepted_idx: usize,
+        /// The log update which the leader applies to its log in order to sync
+        /// with this follower (if the follower is more up-to-date).
+        pub log_sync: Option<LogSync<T>>,
+    }
+
+    /// An update that a replica applies to its log in order to sync with another replica's log.
+    #[derive(Clone, Debug)]
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    pub struct LogSync<T>
```
haraldng, line 65: Move to `util.rs` (next to AcceptedMetaData and similar structs) as it's not a message.

omnipaxos/tests/config/test.toml
```diff
@@ -56,13 +56,10 @@ storage_type = { type = "Memory" }
...
-num_thread = 8
-num_node = 3
-batch_siz = 31
-election_timeout_m = 50
-increment_dela = 2
-num_proposal = 310
-gc_id = 1000
+num_threads = 8
+num_nodes = 3
+batch_size = 1
```
haraldng, line 61: I think we used to use a prime number (31) because that would make checking entries getting decided in batches easier. But I guess if a server gets out of sync in the batches then this check wouldn't work. Anyway, we should use something larger than 1 here to test the batching at least.

omnipaxos_storage/Cargo.toml
```diff
@@ -20,6 +20,6 @@ serde = { version = "1.0", features = ["derive"] }
 commitlog = "0.2.0"
```
haraldng, line 20: remove?

omnipaxos_storage/src/persistent_storage.rs
haraldng, line 1: I'll check this in the next review.

omnipaxos/src/storage.rs
kevin-harrison: All the databases I looked at had different terminology for write batches and transactions. Write batches being what we want here (group of write operations to execute atomically) and transactions being what we specifically designed around not needing (groups of read and write operations where reads may need to some the effects of some writes).

omnipaxos/src/messages.rs
```diff
@@ -73,16 +87,11 @@ pub mod sequence_paxos {
...
-        pub decided_snapshot: Option<SnapshotType<T>>,
-        /// The log suffix.
-        pub suffix: Vec<T>,
-        /// The index of the log where the entries from `suffix` should be applied at (also the compacted idx of `decided_snapshot` if it exists)
-        pub sync_idx: u64,
         /// The decided index
-        pub decided_idx: u64,
-        /// StopSign to be accepted
-        pub stopsign: Option<StopSign>,
+        pub decided_idx: usize,
```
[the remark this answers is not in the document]
kevin-harrison: When we use snapshots the sync_idx will equal the decided_idx the majority of the time. The exception to this is when a stopsign is decided, then sync_idx = decided_idx - 1.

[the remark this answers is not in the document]
kevin-harrison: Not 100% sure what you mean about the Decide message but the follower will adopt the decided index and undecided suffix  of the leader from the AcceptSync, and then reply with an Accepted which the leader will eventually use to see that the undecided entries have become chosen and reply appropriately.

omnipaxos/src/storage.rs
kevin-harrison: Because of the split-brain case, a LogSync with stopsign = None actually means overwrite whatever stopsign you have with None. The LogSync struct is kinda confusing because some fields are diffs between the logs (snapshots when delta and suffix) but some are absolute (decided_idx, accepted_round, stopsign) where you just adopt them even if they are the same as what you already have. Do you think this needs redesigning?

omnipaxos/src/sequence_paxos/follower.rs
kevin-harrison: Do you thinks its preferable to, instead of flushing the batch, just set decided_idx to `max(accdec.dec_idx, self.internal_storage.get_accepted_idx())`?

omnipaxos/src/storage.rs
haraldng: Yes. We're not a database so there is no need to get stuck up on terminology, especially if we're already using "batch" to refer to something else. 

I also disagree that "write batch" is a term that is generally perceived as being atomic (in contrast to a transaction). Therefore the name should include "transaction" or "atomic" in it.

omnipaxos/src/storage.rs
haraldng: I see. I think it's ok for now. But here we can just make this logic more explicit i.e., still do the `match` on the StopSign and in the `None` case we set the stopsign to None instead.

omnipaxos/src/messages.rs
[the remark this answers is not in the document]
haraldng: let's say we have decided up to 100 entries in round 2. The follower only had 40 entries from round 1. Then it should get the suffix of 60 entries that it should apply at index 40 (this used to be called the sync_idx) and the decided _idx should be 100. 

But from your comment it is as I suspected, it will send the decided_idx (i.e., 100) here in a subsequent message.

omnipaxos/src/sequence_paxos/follower.rs
haraldng: I prefer to flush. If it is decided then we might as well flush and make it readable for the users.

omnipaxos/src/sequence_paxos/leader.rs
haraldng: right... I think the usage of function and naming is confusing here.

- In the first if statement: instead of using `self.internal_storage.get_stopsign().is_none()`, use the `accepted_reconfiguration()` function instead.
- I think the `pending_stopsign` field should be renamed to `buffered_stopsign`, as pending can be interpreted as in the progress of being decided.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -16,67 +16,83 @@ where
     pub(crate) fn handle_prepare(&mut self, prep: Prepare, from: NodeId) {
         let old_promise = self.internal_storage.get_promise();
         if old_promise < prep.n || (old_promise == prep.n && self.state.1 == Phase::Recover) {
+            // Flush any pending writes
+            self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG);
```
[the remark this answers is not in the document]
haraldng: On a second thought, let's leave this as it is and we can create an issue and fix it in the future. So we don't add more complexity to this PR which already has many changes.

omnipaxos/src/sequence_paxos/follower.rs
kevin-harrison: I've ended up doing this:
```rust
    pub(crate) fn handle_acceptdecide(&mut self, acc_dec: AcceptDecide<T>) {
        if self.check_valid_ballot(acc_dec.n)
            && self.handle_sequence_num(acc_dec.seq_num, acc_dec.n.pid) == MessageStatus::Expected
        {
            #[cfg(not(feature = "unicache"))]
            let entries = acc_dec.entries;
            #[cfg(feature = "unicache")]
            let entries = self.internal_storage.decode_entries(acc_dec.entries);
            let mut new_accepted_idx = self
                .internal_storage
                .append_entries_and_get_accepted_idx(entries)
                .expect(WRITE_ERROR_MSG);
            let mut new_decided_idx = acc_dec.decided_idx;
            if new_decided_idx > self.internal_storage.get_decided_idx() {
                if new_decided_idx > self.internal_storage.get_accepted_idx() {
                    new_accepted_idx =
                        Some(self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG));
                    new_decided_idx = new_decided_idx.min(new_accepted_idx.unwrap());
                }
                self.internal_storage
                    .set_decided_idx(new_decided_idx)
                    .expect(WRITE_ERROR_MSG);
            }
            if let Some(idx) = new_accepted_idx {
                self.handle_flushed_accepted(acc_dec.n, idx);
            }
        }
    }
```
It feels like we complicating the code for a scenario that should never happen. The leader can never decide entries it hasn't seen itself. So when the follower flushes here it must be the case that its safe to set the `decided_idx` to what the leader told us.

omnipaxos/src/sequence_paxos/follower.rs
kevin-harrison: See above comment

omnipaxos/src/storage.rs
kevin-harrison: I'll do this as a last step in the PR so that you can more clearly see the diff before then.

omnipaxos/src/sequence_paxos/follower.rs
haraldng: I liked the code in the above comment so add it here as well.

omnipaxos/src/sequence_paxos/follower.rs
haraldng: Looks good

omnipaxos/src/storage.rs
```diff
@@ -728,17 +727,23 @@ where
...
-            self.state_cache.stopsign = sync.stopsign.clone();
-            if sync.stopsign.is_some() {
-                self.state_cache.accepted_idx += 1;
+            match sync.stopsign {
+                Some(ss) => {
+                    self.state_cache.stopsign = Some(ss.clone());
+                    self.state_cache.accepted_idx += 1;
+                    sync_txn.push(StorageOp::SetStopsign(Some(ss)));
+                }
+                None => {
```
haraldng, line 736: I think if the stopsign is None already, we can probably skip the setting it again in the storage.

```suggestion
                None if self.state_cache.stopsign.is_some() => {
```

omnipaxos/src/util.rs
```diff
@@ -1,12 +1,29 @@
...
-    messages::sequence_paxos::{LogSync, Promise},
-    storage::{Entry, StopSign},
+    messages::sequence_paxos::Promise,
+    storage::{Entry, SnapshotType, StopSign},
 };
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 use std::{cmp::Ordering, fmt::Debug, marker::PhantomData};
 
+/// An update that a replica applies to its log in order to sync with another replica's log.
```
haraldng, line 10:
```suggestion
/// Struct used to help another server synchronize their log with the current state of our own log.
```

haraldng (changes requested):
Mostly just add more comments/docs. Can now go ahead and refactor the storage file.

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -184,17 +132,22 @@ where
             && self.state.1 == Phase::Accept
             && self.handle_sequence_num(dec.seq_num, dec.n.pid) == MessageStatus::Expected
         {
-            if dec.decided_idx > self.internal_storage.get_accepted_idx() {
-                let new_accepted_idx = self.internal_storage.flush_batch().expect(WRITE_ERROR_MSG);
-                self.handle_flushed_accepted(dec.n, new_accepted_idx);
+            let mut new_decided_idx = dec.decided_idx;
```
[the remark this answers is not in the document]
kevin-harrison: Becomes this:
```rust
    pub(crate) fn handle_acceptdecide(&mut self, acc_dec: AcceptDecide<T>) {
        if self.check_valid_ballot(acc_dec.n)
            && self.handle_sequence_num(acc_dec.seq_num, acc_dec.n.pid) == MessageStatus::Expected
        {
            #[cfg(not(feature = "unicache"))]
            let entries = acc_dec.entries;
            #[cfg(feature = "unicache")]
            let entries = self.internal_storage.decode_entries(acc_dec.entries);
            let mut new_accepted_idx = self
                .internal_storage
                .append_entries_and_get_accepted_idx(entries)
                .expect(WRITE_ERROR_MSG);
            new_accepted_idx = new_accepted_idx.max(self.update_decided_idx_and_get_accepted_idx(acc_dec.decided_idx));
            if let Some(idx) = new_accepted_idx {
                self.reply_accepted(acc_dec.n, idx);
            }
        }
    }
```
Kinda strange imo, but maybe you like it?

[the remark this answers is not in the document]
haraldng: Correct me if I'm wrong, but I don't think the first `new_accepted_idx` can be greater than the second (the one after maybe flushing when deciding)? I think if we remove the `max()` it will be less confusing.

```rust
let mut new_accepted_idx: Option<usize> = self.internal_storage.append_entries_and_get_accepted_idx(...).expect(...);
let flushed_after_decide: Option<usize> = self.update_decided_idx_and_get_accepted_idx(...);
if flushed_after_decide.is_some() {
    new_accepted_idx = flushed_after_decide;
}
if let Some(idx) = new_accepted_idx {
    self.reply_accepted(acc_dec.n, idx);
}
 

```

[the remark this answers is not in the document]
kevin-harrison: It's there because the calculated new_accepted_idx are Options (since we don't want to send Accepted if nothing was flushed). So:
None.max(Some(flush_idx)) = Some(flush_idx) for example.

omnipaxos/src/lib.rs
```diff
@@ -24,6 +24,8 @@ pub use omni_paxos::*;
 
 /// The core replication algorithm of OmniPaxos.
 pub(crate) mod sequence_paxos;
+/// The interface between sequence paxos and the backend storage.
+pub(crate) mod internal_storage;
```
haraldng, line 28: At some point in the future, we will also try to fix some of the redundancies in `StateCache` and `LeaderState`, so I think it's better to further separate the different structs. I prefer we create a `storage` module with:
- `mod.rs` for `Storage`
- `internal_storage.rs` for `InternalStorage`
- `state_cache.rs` for `StateCache`

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -118,6 +112,8 @@ where
             && self.state == (Role::Follower, Phase::Accept)
             && self.handle_sequence_num(acc_ss.seq_num, acc_ss.n.pid) == MessageStatus::Expected
         {
+            // Don't have to handle flushed entries here because we will send Accepted after
```
haraldng, line 115:
```suggestion
            // Flush entries before appending stopsign. The accepted index is ignored here as it will be updated when appending stopsign.
```

omnipaxos/src/sequence_paxos/follower.rs
```diff
@@ -118,6 +112,8 @@ where
...
+            // appending stopsign.
```
haraldng, line 116:
```suggestion
```

omnipaxos/src/storage.rs
```diff
@@ -133,8 +127,11 @@ pub trait Storage<T>
...
     T: Entry,
 {
-    /// Atomically perform and persist all storage operations in order.
-    /// Returns an Error if the transaction was not committed (so it was rolled back).
+    /// **Atomically** perform all storage operations in order.
+    /// For correctness, the operations must be atomic i.e., either all operations are performed
+    /// successfully or all get rolled back. If the `StorageResult` returns as `Err`, the
+    /// operations are assumed to have been rolled back to the previous state before this function
+    /// call.
     fn perform_ops_atomically(&mut self, batch: Vec<StorageOp<T>>) -> StorageResult<()>;
```
haraldng, line 135: Since we don't do any reads, perhaps it would be even more clear for the users if we rename it:

```suggestion
    fn write_atomically(&mut self, ops: Vec<StorageOp<T>>) -> StorageResult<()>;
```

omnipaxos/tests/docs_integration.rs
```diff
@@ -134,18 +133,36 @@ mod docs_integration_test {
...
+        persist_conf.set_path(my_path.to_string()); // set the path of the persistent storage
 
         // Re-create storage with previous state, then create `OmniPaxos`
-        let recovered_storage: PersistentStorage<KeyValue> = PersistentStorage::open(persist_conf);
+        let recovered_storage: PersistentStorage<KeyValue> = PersistentStorage::open(persist_conf); 
         let mut recovered_paxos = omnipaxos_config.build(recovered_storage);
         // END_CODE_EXAMPLE
     }
 
+    // NOTE: We can't import rocksdb options here.
```
haraldng, line 145: remove?

haraldng (changes requested):

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -92,55 +100,158 @@ pub struct PersistentStorage<T>
...
+            match db
+                .get(TRIM)
+                .expect("Couldn't recover storage: Reading compacted_idx failed.")
+            {
+                Some(bytes) => usize::read_from(bytes.as_bytes())
+                    .expect("Couldn't recover storage: Commpacted index has unexpected format."),
+                None => 0,
+            }
+        };
+        drop(log_iter);
```
haraldng, line 153: Add comments for what we're doing here. I guess we are reading and deserializing 8 bytes that corresponds to the key?

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -158,127 +269,152 @@ where
...
+        let mut entries = Vec::with_capacity(to - from);
+        iter.seek(from.to_be_bytes());
+        loop {
+            let entry_bytes = iter.value().ok_or(ErrHelper {})?;
+            entries.push(bincode::deserialize(entry_bytes)?);
+            from += 1;
+            if from == to {
+                break;
+            }
+            iter.next();
```
haraldng, line 344: I prefer using for loops instead of while loops when possible to future-proof against bugs with looping infinitely.

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -92,55 +100,158 @@ pub struct PersistentStorage<T>
...
         ));
-
         Self::open(storage_config)
     }
+
+    /// Get handle to the log column family of the database
+    fn get_log_handle(&self) -> ColumnFamilyRef {
+        self.db
+            .cf_handle(LOG)
+            .expect("Failed to create RocksDB log column family")
```
haraldng, line 176: update error message

omnipaxos_storage/src/persistent_storage.rs
haraldng, line 1: For functions where we currently perform `delete` (mostly the `append_on_prefix `and variations of it): Where it is possible, overwrite existing keys instead of deleting. In LSMs, deletes will write a tombstone so it's like performing an extra write which we should avoid.

haraldng (changes requested):
Overwrite instead of delete where possible in PersistentStorage.

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -199,8 +203,10 @@ where
     }
 
     fn batch_append_on_prefix(&mut self, from_idx: usize, entries: Vec<T>) -> StorageResult<()> {
-        if from_idx < self.next_log_key {
-            let from_key = from_idx.to_be_bytes();
+        // Don't need to delete entries will we overwrite.
```
haraldng, line 206: I don't see where the entries get overwritten? Here we only delete the entries that get dropped (which is correct), but we also need to overwrite entries that are between from_idx and delete_idx

omnipaxos_storage/src/persistent_storage.rs
```diff
@@ -315,8 +321,10 @@ where
     }
 
     fn append_on_prefix(&mut self, from_idx: usize, entries: Vec<T>) -> StorageResult<()> {
-        if from_idx < self.next_log_key {
-            let from_key = from_idx.to_be_bytes();
+        // Don't need to delete entries will we overwrite.
```
haraldng, line 324: same comment as above.

haraldng (changes requested):

haraldng (approved):

haraldng (approved):

soruh:
> All indices are now of type usize instead of u64. This includes the indices used in the Storage trait, Message structs, and Omnipaxos public functions.

I couldn't find the reason for this anywhere. It seems unfortunate to effectively constrain omnipaxos to 64-bit systems.
Using 32-bit ids and assuming 150_000 ops/s (taken from the [EuroSys'23 paper](https://dl.acm.org/doi/pdf/10.1145/3552326.3587441) quoted in the readme). Would give approximately 8 hours until the 32-bit id is exhausted...

Am I missing something obvious / what is the reason for this change?

haraldng:
The change is not performance-related but rather for API reasons, we wanted reading from the omnipaxos log to be similar to the Rust [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html) API. 

Since it is now usize, the user should be able to use it with any compatible system.

soruh:
That makes sense ergonomics wise, my concern would be that a system with a 32bit architecture and one with a 64bit architecture could not be part of the same cluster. Or, for that matter any 32 bit system could exhause the 32bit log index in reasonable time.

PR #128, opened 2023-11-30
Flush batched entries timer

### LINKED ISSUE #127: Flush entries timer
Due to followers batching entries instead of immediately accepting them, its possible for entries to be proposed and never decided (if no further proposals are ever issued). We can add a timer, like the resend message tick timeout, which flushes entries to stop this.

kevin-harrison:

## Issues

Fix #127. `Omnipaxos::tick()` now also flushes batched log entries every `flush_batch_tick_timeout` ticks. This stops entries from being stuck in the batch buffer when there are no incoming proposals.

## Breaking Changes
 - Adds new field `flush_batch_tick_timeout` to `ServerConfig`.

### DIFF
--- a/omnipaxos/src/omni_paxos.rs
+++ b/omnipaxos/src/omni_paxos.rs
@@ -5,7 +5,7 @@ use crate::{
     sequence_paxos::SequencePaxos,
     storage::{Entry, StopSign, Storage},
     util::{
-        defaults::{BUFFER_SIZE, ELECTION_TIMEOUT, RESEND_MESSAGE_TIMEOUT},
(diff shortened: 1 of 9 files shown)


kevin-harrison:
> Looks good to me. I just want to confirm with you that we don't need to check that we're in the accept phase to flush these entries? Because they get buffered up in memory only in the accept phase.

Yes exactly the entries will get flushed on entering the Prepare phase and will only start batching up in the Accept phase. But to be safe its probably better to only flush in the accept phase to prevent future bugs. I'll make that change now.

haraldng (approved):
Thanks!

PR #141, opened 2024-11-05
expose API for accepted phase

Shubham8287:
Added a method to expose `Phase` of an omnipaxos node. 
One usecase for us is to be able to recognise that elected leader is now ready for routing requests.

> [] You ran the local CI checker with `./check.sh` with no errors 

a `omnipaxos/tests/ble_test` seems broken in master?

### DIFF
--- a/omnipaxos/src/omni_paxos.rs
+++ b/omnipaxos/src/omni_paxos.rs
@@ -421,7 +421,7 @@ where
 
         ui::OmniPaxosStates {
             current_ballot: self.ble.get_current_ballot(),
-            current_leader: self.get_current_leader(),
+            current_leader: self.get_current_leader().map(|(leader, _)| leader),
             decided_idx: self.get_decided_idx(),
             heartbeats: self.ble.get_ballots(),
             cluster_state,
--- a/omnipaxos/src/omni_paxos.rs
+++ b/omnipaxos/src/omni_paxos.rs
@@ -2,7 +2,7 @@ use crate::{
     ballot_leader_election::{Ballot, BallotLeaderElection},
     errors::{valid_config, ConfigError},
     messages::Message,
-    sequence_paxos::SequencePaxos,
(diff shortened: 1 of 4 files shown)


omnipaxos/src/omni_paxos.rs
```diff
@@ -421,6 +421,11 @@ where
             cluster_state,
         }
     }
+
+    /// Returns whether Sequence Paxos instance is in Accept Phase.
+    pub fn is_accepted(&self) -> bool {
+        self.seq_paxos.get_state().1 == Phase::Accept
+    }
```
haraldng, line 428: I think it would be better to extend the [get_current_leader()](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/omni_paxos.rs#L272) function to return `Option<(NodeId, Phase)>`

haraldng (changes requested):

omnipaxos/src/omni_paxos.rs
Shubham8287: As I understand it, an Omnipaxos instance is only aware of its own `Phase`, not the leader’s. Returning `Phase` in the `get_current_leader` method may imply that `Phase` represents the leader, but that’s not accurate. Isn't?

haraldng: If the follower is in the `Accept` phase then that also applies to the leader. So one solution could be to change the return type to `Option<(NodeId, bool)>` and only return `true` if our own phase is `Accept`.

Shubham8287: Made changes as suggested, also added a doc comment, please feel free to suggest any changed to make that clearer.

haraldng (changes requested):
It looks good. Could you please update the tests where the function is called so that they compile and CI can run? Thank you for contributing! you can check that things run with `./check.sh`

haraldng (approved):

PR #144, opened 2024-12-06
Support for single peer cluster

### LINKED ISSUE #143: Is it possible to run Omnipaxos with a single peer?
I noticed that in the `ClusterConfig::validate` function you assert that there are more than 1 peer. Is there a strict requirement why an Omnipaxos cluster with a single node wouldn't work?

tillrohrmann:

## Issues

Fix #143.

## Changes

This commit enables OmniPaxos to run with a single peer. What needed to
be changed is to call the proper handle methods instead of directly
updating the LeaderState with one's own promise and accepted index. By
using the handle methods, we ensure that all required actions (e.g. updating
the decided index, becoming a leader) are properly triggered.

These changes were tested by running the following tests with a single node:

* ble_test
* consensus_test
* unicache_test
* flexible_quorum_test
* trim_test
* batching_test

The other integration tests require more than 1 peer in order to test follower
behavior.

### DIFF
--- a/examples/dashboard/src/util.rs
+++ b/examples/dashboard/src/util.rs
@@ -51,7 +51,8 @@ pub(crate) fn parse_arguments() -> Result<(u64, u64, Duration, u64), String> {
             }
             "duration" => {
                 i += 1;
-                duration_in_seconds = args[i].parse().expect("Invalid duration."); // Set the default duration to 10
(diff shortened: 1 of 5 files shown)


haraldng:
We should add this feature when we have implemented bootstrapping, see #143 .

PR #155, opened 2025-02-10
Fix `set_priority` to trigger election

Shubham8287:
As I understand from the comment on [set_priority](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/omni_paxos.rs#L399), setting a follower with a higher priority than the current leader is expected to trigger an election. However, this does not seem to happen in practice.

Changes I made in the PR that resolve this:

- `self.current_ballot` was not getting updated unless node becomes unhappy [(see here)](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/ballot_leader_election.rs#L269). This makes any "restarted" node to have lower epoch than its leader.

- Modified `update_leader` to consider `self.current_ballot` when updating `self.leader` .

These changes are based on my understanding of a node's priority, which is:
"In a cluster, a leader should not have a lower priority than any of its followers."
Let me know, if it is not true.
 
### DIFF
--- a/omnipaxos/src/ballot_leader_election.rs
+++ b/omnipaxos/src/ballot_leader_election.rs
@@ -208,6 +208,12 @@ impl BallotLeaderElection {
             self.leader = seq_paxos_promise;
             self.happy = true;
         }
+
+        // Sync epoch with leader
+        if self.leader.n > self.current_ballot.n {
+            self.current_ballot.n = self.leader.n;
+        }
+
         if self.leader == self.current_ballot {
             Some(self.current_ballot)
         } else {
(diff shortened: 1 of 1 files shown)


haraldng:
The priority was initially implemented for the purpose to have higher priority to become leader **when** an election occurs. 

Would the [`try_become_leader()`](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/omni_paxos.rs#L391) be more suitable for your use case?

haraldng:
This change could potentially cause unnecessary elections because the ballot is increased without checking quorum connectivity (QC).

For example:
1. Node A is a QC, stable leader (and is the only QC node in the cluster)
2. Node B is not QC but given higher priority than A.
3. With the changes B will eventually gain leadership (because of priority) and eventually become unhappy because it's not QC.
4. Node A will eventually take leadership because B is unhappy and A is QC.
Steps 3 and 4 might be repeated endlessly.

I think the [try_become_leader()](https://github.com/haraldng/omnipaxos/blob/master/omnipaxos/src/omni_paxos.rs#L391) is a better option for now.

PR #161, opened 2025-07-10
Fix CI

kim:
Attempt to fix GHA


kim:
Sorry, wrong upstream

PR #170, opened 2026-09-17
Async Runtime

### LINKED ISSUE #60: Async API
We would like to have an async API e.g. `append_notify()` such that when an entry is decided it can be notified to the user via a future/one shot channel.

Ideally we would want a runtime agnostic API, but we probably have to commit to tokio right now. The API should also be optional using a feature flag

haraldng:

## Issues

Fix #60

## Breaking Changes

None for existing users. Changes to the `omnipaxos` crate are additive; the new `async_runtime` feature is opt-in (off by default).

- `Message<T>` gains an `AsyncRuntime(...)` variant, but only when the `async_runtime` feature is enabled. Exhaustive matches on `Message<T>` that also enable `async_runtime` need a new arm (or a wildcard).
- New unconditional (not feature-gated) public methods on `OmniPaxos`: `get_accepted_idx()`, `get_batched_len()`, and `get_pid()` — additive getters only.

## Other Changes

- New `omnipaxos_runtime` crate: `spawn_actor` + cloneable `OmniPaxosHandle` (`Send`/`Sync`) with async APIs.
- Runtime-agnostic `AsyncRuntime` trait; `TokioRuntime` behind `tokio_runtime`. Transport is not owned by the crate — callers drain `outgoing_messages()` and feed `handle_incoming()`.
- `append_notify` is the main new capability: append and `.await` the decided log index from any node (followers route to the leader via `TaggedProposal` / `Assigned`). Resolves with `Ok(idx)` or `AppendError::{Timeout, Superseded { log_entry_idx }, Propose, Shutdown, TooManyOutstanding}`.
- All outgoing egress goes through a shared non-blocking `try_send` + bounded internal buffer (`max_outgoing_buffered` / channel capacity). Oldest buffered messages are dropped under sustained backpressure, so a slow consumer cannot OOM the actor or stall ticks/election/commands.
- New `examples/kv_store_async` end-to-end demo and `omnipaxos_runtime/README.md`.
- `omnipaxos`: new `async_runtime` feature (pulls in `uuid`); `serde` wired so it can chain uuid’s serde support; `test_all_features.sh` updated for the new features.
- Tests: 24 integration tests in `omnipaxos_runtime/tests/async_runtime_test.rs` covering leader/follower notify routing, Prepare-phase waiting, timeouts, supersession, batching (local append + forwarded proposal), subscriber backpressure / limits, reconfiguration / shutdown error propagation, and full-outgoing-channel liveness (core egress and `TaggedProposal`/`Assigned`).
- Incidental clippy/fmt cleanups needed for `./check.sh`: lifetime elision / `needless_return` in `omnipaxos_storage`, `omnipaxos_ui`, and `omnipaxos::unicache`, plus a stray non-feature entry removed from `omnipaxos/test_all_features.sh`.

### Follow-ups (not in this PR)

- OmniPaxos book docs for the async runtime.


omnipaxos_runtime/src/actor.rs
```diff
@@ -0,0 +1,754 @@
...
+            return;
+        }
+        // Read once, from the trailing sub's position. `entries[i]` is the log
+        // entry at position `min_next + i`.
+        let Some(entries) = self.op.read_decided_suffix(min_next) else {
+            return;
+        };
+        // Fan out. Preserve the drain-and-refill idiom so closed subs are
+        // pruned without borrow-checker gymnastics.
+        let subs = std::mem::take(&mut self.decided_subscribers);
```
haraldng: most of the time, subscribers should still be around so might not be efficient to `mem::take` and then push again.

Shubham8287:
This looks exciting. @haraldng do you have plans to make `Storage` trait async too? as this PR looks like a step in that direction. We would love to have that, and even open contribute it as well.

haraldng:
> This looks exciting. @haraldng do you have plans to make `Storage` trait async too? as this PR looks like a step in that direction. We would love to have that, and even open contribute it as well.

@Shubham8287 I would love to support async storage, but that would probably require quite big changes in the core `omnipaxos`code. As of right now, I don't have concrete plans on how to do it but would be happy to chat about it if you would be interested in prototyping it.

